PR #9, opened 2017-05-18
option to bring-your-own-binding

mckaymatt:
@fafhrd91 What do you think about providing an option to avoid using rust-cpython or pyo3? 
Are you even the right project to talk to? I know nothing about PyO3 but it just seemed like this was the most actively developed Setuptools plugin for Rust. 

I'm making a cookiecutter project to enable cross-platform build and publish of Python wheels that have rust extensions and I used CFFI instead of those two. I needed to make this fork to unblock myself.

The new Cookiecutter is here https://github.com/mckaymatt/cookiecutter-pypackage-rust-cross-platform-publish and example output is here https://github.com/mckaymatt/trust_pypi_example specifically here https://github.com/mckaymatt/trust_pypi_example/blob/master/requirements_dev.txt#L15

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ Cargo.lock
 /example/dist/
 /example/extensions/target/
 /example/*.egg-info
+.idea/
+
--- a/README.rst
+++ b/README.rst
@@ -73,7 +73,8 @@ You can define rust extension with `RustExtension` class:
    :param bool debug: Controls whether --debug or --release is passed to cargo. If set to
                       None then build type is auto-detect. Inplace build is debug build
                       otherwise release. Default: None
-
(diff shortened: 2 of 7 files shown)


mckaymatt:
cc @cathalgarvey

fafhrd91:
I think "no-biding" is good idea. but could you make one argument `binding` and enum(or just consts) like NoBinding, PuO3, RustCPython, otherwise extra argument adds ambiguity

cathalgarvey:
Hey, thanks for the heads up about the cookiecutter project, will look forward to reading through it! I've had trouble uploading my wheels to PyPI because of the filenames, apparently PyPI sometimes doesn't like "x86_64"..

fafhrd91:
@mckaymatt what would be really cool, if your cookiecutter could generate packages for different bindings

fafhrd91:
close

PR #11, opened 2017-06-14
binding options as an enum

mckaymatt:
how about this @fafhrd91 
see  https://github.com/PyO3/setuptools-rust/pull/9

I can get my package to work with these changes, but I can't get the example in the repo to work. It doesn't look like you actually pull in the PyO3 bindings in the .toml file.

### DIFF
--- a/setuptools_rust/check.py
+++ b/setuptools_rust/check.py
@@ -55,11 +55,12 @@ class check_rust(Command):
                     "Can not file rust extension project file: %s" % ext.path)
 
             features = set(ext.features)
-            features.update(cpython_feature(pyo3=ext.pyo3))
+            features.update(cpython_feature(binding=ext.binding))
 
             # build cargo command
-            args = (["cargo", "check", "--lib", "--manifest-path", ext.path,
-                     "--features", " ".join(features)]
+            feature_args = ["--features" + " ".join(features)] if features else []
+            args = (["cargo", "rustc", "--lib", "--manifest-path", ext.path]
+                    + feature_args
                     + list(ext.args or []))
 
             # Execute cargo command
--- a/setuptools_rust/test.py
+++ b/setuptools_rust/test.py
@@ -57,17 +57,18 @@ class test_rust(Command):
                     "Can not file rust extension project file: %s" % ext.path)
 
             features = set(ext.features)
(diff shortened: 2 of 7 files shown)


fafhrd91:
lgtm

fafhrd91:
I'll check example package agains this change

setuptools_rust/utils.py
```diff
@@ -6,15 +6,29 @@
 import semantic_version
 
 
-def cpython_feature(ext=True, pyo3=False):
+class Binding:
```
messense: Since `setuptools-rust` supports Python 2 & 3, better use `class Binding(object):` to get ride of old style class behavior even though that don't really matter here. :-)

setuptools_rust/test.py
```diff
@@ -57,17 +57,18 @@ def run(self):
...
 
             features = set(ext.features)
-            features.update(cpython_feature(ext=False, pyo3=ext.pyo3))
+            features.update(cpython_feature(ext=False, binding=ext.binding))
 
             # build cargo command
-            args = (["cargo", "test", '--color', 'always', "--manifest-path",
-                     ext.path, "--features", " ".join(features)]
+            feature_args = ["--features " + " ".join(features)] if features else []
+            args = (["cargo", "rustc", "--lib", "--manifest-path", ext.path]
```
messense: Is it used to run `cargo test`, why it runs `cargo rustc` now?

setuptools_rust/check.py
```diff
@@ -55,11 +55,12 @@ def run(self):
...
 
             features = set(ext.features)
-            features.update(cpython_feature(pyo3=ext.pyo3))
+            features.update(cpython_feature(binding=ext.binding))
 
             # build cargo command
-            args = (["cargo", "check", "--lib", "--manifest-path", ext.path,
-                     "--features", " ".join(features)]
+            feature_args = ["--features" + " ".join(features)] if features else []
+            args = (["cargo", "rustc", "--lib", "--manifest-path", ext.path]
```
messense: `cargo check` ?

setuptools_rust/test.py
mckaymatt: good catch. This was a sloppy copy and paste.

fafhrd91: @mckaymatt could you update your pr?

PR #16, opened 2017-07-31
Add optional option to RustExtension

### LINKED ISSUE #15: Support optional option
Just like distutils:

https://docs.python.org/3/distutils/setupscript.html#other-options

> The `optional` option is a boolean; if it is true, a build failure in the extension will not abort the build process, but instead simply not install the failing extension.

messense:
Fixes #15


messense:
Sure. You want the new version be `0.7.0` or `0.6.2`?

fafhrd91:
0.6.2

PR #28, opened 2018-02-24
Improve compatibility of `build_rust` with `build_ext`

### LINKED ISSUE #27: Patched commands ignore `setup.cfg` defined options
Hi there,
the `setup.cfg` file allows to define default values for arguments of setuptools / distutils commands. With the following file:
```toml
[build_rust]
inplace = true
release = true
```
simply running `python setup.py build_rust` will build all the extensions *inplace* and in release mode as expected.

Unfortunately, running `python setup.py build` or `python setup.py build_ext` does not. I suspect there's a step in the monkey-patching where the `build_rust` command is ran without having `finalize_options` called or something like that.

althonos:
### Rationale

The behaviour of `build_ext` is the following: when generating compilation 
artifacts (such as `.o` files), it will put them in a directory defined by the 
`build_temp` argument, and within that directory create a directory named after 
the project name in `setup.py`, and them dump its temporary files there.

The only way to do that with `build_rust` was through editing the 
`CARGO_TARGET_DIR` environment variable, but this is not practical and also does 
not comply well with the different available configuration files (`setup.cfg`, 
user `distutils.cfg` and the like).

### Implementation

* if the user gives the `--build-temp` argument to the CLI, or sets `build_temp` 
within the `[build]`, `[build_ext]` or `[build_rust]` sections of any 
configuration file, then cargo will use that directory as a target directory.
* if no directory is set, then cargo will use the default `build_temp` (which is 
`build/temp.<platform>-<version>`).
* if `CARGO_TARGET_DIR` is set, then it will overwrite any python-defined setting 
and build to that directory directly.

I also made `build_rust` inherit from `build_ext` the following options: `inplace` 
and `debug`.

### Perks

* if a library has more than one extension modules with `Cargo.toml` files in 
different subdirectories, cargo would create a local target directory for each 
library, and thus rebuild several times the same cargo modules (for instance, 
`PyO3` !)
* removes the need for a `clean_rust` command, since `python setup.py clean` 
removes the `build_temp` directory
* fixes #27 :wink: 
* fixes the following error: when `CARGO_TARGET_DIR` is set, and multiple extensions are built for a project, the wrong library can be copied (since `build_rust` did not check for the library name beforehand)

### Drawbacks

* not really backwards compatible

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -28,6 +28,8 @@ class build_rust(Command):
          "Force debug to false for all rust extensions "),
         ('qbuild', None,
          "Force enable quiet option for all rust extensions "),
+         ('build-temp', 't',
+          "directory for temporary files (cargo 'target' directory) "),
     ]
     boolean_options = ['inplace', 'debug', 'release', 'qbuild']
 
@@ -37,10 +39,18 @@ class build_rust(Command):
         self.debug = None
         self.release = None
         self.qbuild = None
+        self.build_temp = None
 
     def finalize_options(self):
         self.extensions = [ext for ext in self.distribution.rust_extensions
                            if isinstance(ext, RustExtension)]
+
+        # Inherit settings from the `build_ext` command
+        self.set_undefined_options('build_ext',
+            ('build_temp', 'build_temp'),
+            ('debug', 'debug'),
+            ('inplace', 'inplace'),
+        )
 
     def build_extension(self, ext):
         executable = ext.binding == Binding.Exec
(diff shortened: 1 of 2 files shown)


adrienball:
+1

I am running into troubles with a rust extension generating multiple `.dylib` files in the target dir. The following piece of code in `setuptools_rust` picks the first binary matching the wildcard, which varies across platforms:

```
            if sys.platform == "win32":
                wildcard_so = "*.dll"
            elif sys.platform == "darwin":
                wildcard_so = "*.dylib"
            else:
                wildcard_so = "*.so"

            try:
                dylib_paths.append(
                    (ext.name, glob.glob(
                        os.path.join(target_dir, wildcard_so))[0]))
            except IndexError:
                raise DistutilsExecError(
                    "rust build failed; unable to find any %s in %s" %
                    (wildcard_so, target_dir))
```

fafhrd91:
awesome! thanks

fafhrd91:
@althonos do you want me to make new release?

althonos:
@fafhrd91: yes, that would be nice of you, I have a project that would likely benefit from it !

I'm also working on an automatic `Cargo.toml` generation for extensions based on source files and package metadata, would you be interested in a PR to integrate it or should I rather work on it as an external package ?

fafhrd91:
Auto generating cargo toml seems interesting idea

fafhrd91:
new version is released

Issue #40, opened 2018-09-07
license?


craftyguy:
Hi! I would like to package this for Arch Linux's AUR, but there is no clear indication of the license this software is released under. PyPi lists it as MIT but I just want to confirm.

marcelotrevisani:
There is a request in ``conda-forge`` as well and we need to know what is the license of this project. If possible, please add the license into the project and add in``MANIFEST.in``

marcelotrevisani:
cc: @konstin

craftyguy:
FYIY, the [setup.py does mention the MIT license](https://github.com/PyO3/setuptools-rust/blob/master/setup.py#L17), but it would be great to have the license explicitly included in MANIFEST.in or a LICENSE file.

marcelotrevisani:
I opened a PR for that
https://github.com/PyO3/setuptools-rust/pull/41

craftyguy:
Awesome, thanks @marcelotrevisani !

PR #41, opened 2018-09-08
Add license file

marcelotrevisani:
Add LICENSE file to the project

Please review @fafhrd91

Related to https://github.com/PyO3/setuptools-rust/issues/40

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017-2018 PyO3 project & contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2017-2018 Nikolay Kim & contributors
```
messense: I think it's better to add `MIT License` in the first line.

marcelotrevisani: Got it

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017-2018 Nikolay Kim & contributors
```
fafhrd91: Lets use PyO3 project and contributors

marcelotrevisani: Got it! I'm going to change

PR #45, opened 2018-11-05
Fix CI failing because of Cargo.toml format change

althonos:
Rust used to support crate names in `Cargo.toml` containing dots, but doesn't anymore. This means the extension name cannot be used as a crate name, so it is escaped with dashes instead of dots (unless someone has a better idea ?).

I also patched `example_tomlgen/Cargo.toml` to use the git version of PyO3 instead of the v4 which does not build successfully.

### DIFF
--- a/example_tomlgen/setup.cfg
+++ b/example_tomlgen/setup.cfg
@@ -10,7 +10,7 @@ zip_safe = false
 create_workspace = true
 
 [tomlgen_rust.dependencies]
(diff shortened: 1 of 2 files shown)


althonos:
> Thanks this looks like the most viable solution.

The other solution would be to use relative name (e.g. `english` instead of `hello_english` for `hello.english`).

> I'd prefer using a pyo3 0.5.0-alpha.3 instead of git though

I'll be patching that.

PR #53, opened 2019-03-28
Remove python 2 support

konstin:
This also allows removing the python3 feature from pyo3


konstin:
oh thanks, I totally missed that. Should be fixed now, let's see what travis ci says

PR #62, opened 2020-03-11
Tentative Cygwin support for dll

KoichiYasuoka:
Quick hack to support Cygwin-based dll.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -282,6 +282,7 @@ class build_rust(Command):
                     except subprocess.CalledProcessError as e:
                         pass
 
+            # executables and win32(cygwin)-dll's need X bits
             if executable or sys.platform == "win32" or sys.platform == "cygwin":
                 mode = os.stat(ext_path).st_mode
                 mode |= (mode & 0o444) >> 2  # copy R bits to X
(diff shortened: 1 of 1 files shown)


setuptools_rust/build.py
```diff
@@ -282,7 +282,7 @@ def build_extension(self, ext):
                     except subprocess.CalledProcessError as e:
                         pass
 
-            if executable:
+            if executable or sys.platform == "win32" or sys.platform == "cygwin":
```
kngwyu: You mean we need `chmod` for dlls?

kngwyu:
Thanks!

setuptools_rust/build.py
KoichiYasuoka: Yes. For example, [tokenizers.dll](https://pypi.org/project/tokenizers/) needs to `chmod u+x` when installing on Cygwin.

kngwyu: - How about doing this hack only for cygwin? Is it also required for win32 platform?
- Please add a comment that explains why cygwin needs it.

KoichiYasuoka: This is for "cross-compile" between win32 and cygwin. Now I'm trying to compile PyO3 dll's for win32 with cygwin gcc, and then make wheels for win32 with setuptools_rust. In such cases, I need `sys.platform == "win32"` here. I agree that it is very rare case now.

KoichiYasuoka: I think that MSVC always outputs dll's with `-rwxr-x---` modes, so they don't need `chmod` here explicitly.

kngwyu:
LGTM, thank you!

PR #65, opened 2020-06-19
Moved the metadata into setup.cfg

KOLANICH:


althonos:
Thanks @KOLANICH !

KOLANICH:
BTW in some future yet another transition may be needed: https://www.python.org/dev/peps/pep-0621/

althonos:
I am aware of PEP621 and the mess created by PyPA when they decided to introduce yet another file when `setup.cfg` was becoming a standard (at least it was supported by a bunch of tools like `flake8`, `coverage`, etc. for configuration). However, they probably won't deprecate a `setuptools` install before a long time, since transitioning old packages will take time. By just declaring the build backend to be `setuptools` I think we are safe for some time :wink:

PR #66, opened 2020-07-25
Update html-py-ever to pyo3 0.11.1

davidhewitt:
Turns out to be easy enough.

`kuchiki::NodeRef` is not `Send`, so had to use `#[pyclass(unsendable)]`.


kngwyu:
Looks `example_tomlgen` causes an error on CI.

PR #67, opened 2020-07-25
Update instructions to manylinux2014

davidhewitt:
The `manylinux1` image has an extremely old glibc, which means that the image doesn't even run on many systems! https://mail.python.org/pipermail/wheel-builders/2016-December/000239.html

I propose we update to `manylinux2014` images, which use slightly newer package versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ export PATH="$HOME/.cargo/bin:$PATH"
 
 cd /io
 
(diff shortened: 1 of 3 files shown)


kngwyu:
I think both `manylinux1` and `manylinux2014` settings should be noted.
`manylinux1` is not dead and still used by many packages like [numpy](https://pypi.org/project/numpy/#files).

davidhewitt:
👍 added a note at the end saying it's possible to replace `manylinux2014` with the other choices.

kngwyu (approved):

kngwyu:
Thanks!

PR #86, opened 2020-11-07
Avoid f-strings for now

### LINKED ISSUE #85: New release breaks on python 3.5
The new release of setuptools-rust isn't compatible with python 3.5 since it uses fstrings. This is fine, especially since 3.5 is EoL now, but there is no minimum python version set in the package metadata. This means that people trying to use python 3.5 with setuptools-rust it will have pip install the new package version and fail. For example:

https://github.com/Qiskit/retworkx/pull/189/checks?check_run_id=1347529773#step:6:30

It would be good if the next release marked that the minimum supported version was 3.6.

mtreinish:
This commit removes the use of f strings from tomlgen module. The package
metadata didn't set a minimum python version and fstrings don't work
with python 3.5. To enable having the last release of setuptools-rust
prior to setting the python_requires work with python 3.5 we need to
have a release which will work for 3.5. With this change setuptools-rust
should work fine with Python 3.5. So we can release this and then set
the minimum supported python version to 3.6 and reintroduce the use of
fstrings.

Fixes #85

### DIFF
--- a/setuptools_rust/tomlgen.py
+++ b/setuptools_rust/tomlgen.py
@@ -188,7 +188,7 @@ class tomlgen_rust(setuptools.Command):
         for section in sections:
             if self.cfg.has_section(section):
                 for dep, options in self.cfg.items(section):
-                    yield dep, toml.loads(f"{dep} = {options}")[dep]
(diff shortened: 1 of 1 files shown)


davidhewitt:
Thanks!

PR #90, opened 2020-11-14
Use target CARGO_BUILD_TARGET if specified

Anthchirp:
We ran into a [build issue on conda-forge](https://github.com/conda-forge/tokenizers-feedstock/pull/8#issuecomment-723472759) recently, where setuptools-rust does not pick up that a CARGO_BUILD_TARGET environment variable is set. The effect of this is that the rust compilation step succeeds, but the build artifacts land in a `.../target/x86_64-unknown-linux-gnu/release/...` path. `setuptools-rust` still looks for them in the `.../target/release/...` parth though, which the fails the build.

This change might fix the issue. I haven't tested it though, and I'm not a rust developer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+ - Respect `CARGO_BUILD_TARGET` environment variable if set. [#90](https://github.com/PyO3/setuptools-rust/pull/90)
+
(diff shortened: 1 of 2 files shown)


davidhewitt:
👍 thanks! I've pushed a changelog entry & will merge now.

You should be able to test it by installing `setuptools-rust @ git+https://github.com/PyO3/setuptools-rust.git` using pip.

Issue #91, opened 2020-12-10
Please include version in setup.cfg


tiran:
I'm currently in the process to package setuptools-rust for Fedora Linux. There is a small, easy to fix problem in packaging of setuptools-rust. ``setup.cfg`` does not include version number of the package. The RPM packaging build chain re-creates the egg-info directory based on information from ``setup.py`` and ``setup.cfg``. Without a version entry in setup.cfg, setuptools assumes version ``0.0.0``.

```
$ rpmbuild -bb python-setuptools-rust.spec
...
RPM build errors:
    File not found: /BUILDROOT/python-setuptools-rust-0.11.5-1.fc32.x86_64/usr/lib/python3.8/site-packages/setuptools_rust-0.11.5-py3.8.egg-info
$ ls /BUILDROOT/python-setuptools-rust-0.11.5-1.fc32.x86_64/usr/lib/python3.8/site-packages/
setuptools_rust  setuptools_rust-0.0.0-py3.8.egg-info
```

* https://packaging.python.org/guides/distributing-packages-using-setuptools/#version
* https://packaging.python.org/guides/distributing-packages-using-setuptools/#choosing-a-versioning-scheme

tiran:
The change may not be required after all. I have tracked down the issue to setuptools. The build failure might be caused by a bug in setuptools < 42 and an unsufficient version requirement in setuptools_scm. My work machine is on Fedora 32 with setuptools 41.6.

I have filed https://github.com/pypa/setuptools_scm/pull/504 with setuptools_scm and requested a version bump to setuptools >= 42.

althonos:
No, I still think it's a good idea to have the version in `setup.cfg`, plus maybe somewhere in `__init__.py`. Starting with `setuptools >= 46.1` we can write the version in `__init__.py` and use `attr: setuptools_rust.__version__` in `setup.cfg`.

tiran:
That sounds like a great idea!

Ronny has merged my PR to bump minimum setuptools versions to 42. He will release a new version of setuptools_scm next week.

I can also confirm that my issue was caused by old setuptools and setuptools_scm. The version field is autofilled with setuptools_scm 4.1.2 and setuptools 49 of Fedora 33.

tiran:
I have updated my PR #93 to use ``version = attr: setuptools_rust.__version__``.

PR #93, opened 2020-12-10
Add version information to setup.cfg

tiran:
- add `setuptools_rust.__version__`
- populate setup.cfg's version field from `setuptools_rust.__version__`
- require setuptools >= 46.1

Version ``0.11.6.dev1`` means that the current main branch is in
development phase of an upcoming 0.11.6 release.

Resolves: https://github.com/PyO3/setuptools-rust/issues/91

### DIFF
--- a/setuptools_rust/__init__.py
+++ b/setuptools_rust/__init__.py
@@ -8,6 +8,8 @@ from .test import test_rust
 from .tomlgen import tomlgen_rust, find_rust_extensions
 from .utils import Binding, Strip
 
+__version__ = "0.11.6.dev1"
+
 __all__ = (
     "RustExtension",
     "Binding",
--- a/setuptools_rust/__init__.py
+++ b/setuptools_rust/__init__.py
@@ -7,8 +7,8 @@ from .extension import RustExtension
 from .test import test_rust
 from .tomlgen import tomlgen_rust, find_rust_extensions
 from .utils import Binding, Strip
+from .version import version as __version__
 
-__version__ = "0.11.6.dev1"
 
 __all__ = (
     "RustExtension",
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@ build
 target
 .pytest_cache
 *.egg-info
+setuptools_rust/version.py
 pyo3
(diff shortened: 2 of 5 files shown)


setuptools_rust/__init__.py
```diff
@@ -8,6 +8,8 @@
 from .tomlgen import tomlgen_rust, find_rust_extensions
 from .utils import Binding, Strip
 
+__version__ = "0.11.6.dev1"
```
davidhewitt, line 11: To make it easier to do releases, can we use `setuptools_scm` version hard-coding?

e.g. in `pyproject.toml`:

```
[tool.setuptools_scm]
write_to = "pkg/version.py"
```

and then refer to the generated file here?

davidhewitt (commented):
Thanks for this PR!

The purpose of `setuptools_scm` is to keep the release process simple without manually updated versions, so it'd be great if we can somehow leverage that functionality still.

See my other comment...

setuptools_rust/__init__.py
tiran: Nice trick! I was not aware of this feature. Please check out the update to my PR.

davidhewitt: Thanks! I will test this quickly 👍

setuptools_rust/__init__.py
```diff
@@ -7,6 +7,8 @@
 from .test import test_rust
 from .tomlgen import tomlgen_rust, find_rust_extensions
 from .utils import Binding, Strip
+from .version import __version__
```
davidhewitt, line 10: From testing I think this needs to be:

```suggestion
from .version import version as __version__
```

... does that match your experience too?

tiran: Oops, I initially used a template with ``__version__``. I removed the template and forgot to re-generate ``version.py`` before I pushed the change.

davidhewitt: 👍 yeah I considered adding a template but thought the default was nicer too.

davidhewitt:
Many thanks! CI seemed to have an issue so I force-pushed to fix; hope that's okay.

PR #96, opened 2020-12-22
Enable PYTHON_SYS_EXECUTABLE

KoichiYasuoka:


davidhewitt:
Thanks for the PR! Can you please add a little detail in what situation you need this functionality?

As I understand, with setuptools-rust we are installing the Rust extension to be run with `sys.executable`. I worry that enabling these environment variables could cause installations to be built incorrectly.

KoichiYasuoka:
I often use `setuptools-rust` on Cygwin, where the `sys.executable` is `/usr/bin/python3.7`. But in `rust` (Windows) view the executable file is `C:\cygwin64\bin\PYTHON~1.EXE`, so I use `PYTHON_SYS_EXECUTABLE` to tell `rust` where the executable file is.

davidhewitt:
Ah, I see - that makes sense. It's a bit of a weird case; I agree that what you're doing is the right solution.

FYI these days I use WSL 2 instead of Cygwin on Windows, which I find works very very nicely and is a bit friendlier to use than Cygwin 😄 

I'm going to just push a CHANGELOG entry to this branch, and then I'll merge. Thanks very much for your contribution!

Issue #51, opened 2019-03-08
html-py-ever: travis -> pypi


clbarnes:
It would be very useful to see an example of how to deploy python/rust wheels to PyPI from travis.

I imagine it's a fairly simple frankensteining of the current html-py-ever and [PyPA's manylinux demo](https://github.com/pypa/python-manylinux-demo) but a concrete example would be great.

clbarnes:
Here's an example of a repo with a travis config which builds and uploads wheels for pythons 3.5-3.7, and uploads a source distribution as well: https://github.com/pattonw/rust-pyn5 (https://pypi.org/project/pyn5/#files)

PR #100, opened 2021-01-14
github workflows for html-py-ever

0kalekale:
as per issue #51 , added a example travis ci config which builds uploads wheels for pythons 3.6-3.8 and uploads a source distribution to pypi.


davidhewitt:
Thanks! Personally I prefer github actions these days, I wonder if you've tried that? I can help write an example for that if you're interested.

0kalekale:
yeah i too prefer github action, infact just a while ago i used it to build and test my own python-rust module.

https://www.github.com/0kalekale/ipaddr/tree/master/.github%2Fworkflows%2Ftest-module.yml

the issue happen to mention travis so, what do you say?

davidhewitt:
👍 awesome! I think if we provide a github actions example, that's probably better in my opinion so we could close the issue with that?

0kalekale:
sure

0kalekale:
@davidhewitt with this you can close the issue

davidhewitt:
👍 thank you so much for this!

Issue #101, opened 2021-01-14
setuptools is not listed as installation requirement 


tiran:
setuptools_rust has a runtime dependency on setuptools. However ``setup.cfg`` does not include setuptools as a requirement. While pip-based installations and virtual environments always have setuptools, other package managers do not necessarily install setuptools.

For example Fedora's RPM macros automatically create dependencies from package metadata. The auto-generate RPM package for setuptools-rust 0.11.6 is missing a runtime dependency on setuptools.

PR #102, opened 2021-01-14
Add runtime dependency on setuptools

tiran:
setuptools_rust has a runtime dependency on setuptools. Ensure that all
installations of setuptools_rust have a sufficiently recent version of
setuptools.

Fixes: https://github.com/PyO3/setuptools-rust/issues/101

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ classifiers =
 [options]
 packages = setuptools_rust
 zip_safe = True
-install_requires = semantic_version>=2.6.0; toml>=0.9.0
+install_requires = setuptools>=46.1; semantic_version>=2.6.0; toml>=0.9.0
 setup_requires = setuptools>=46.1; wheel; setuptools_scm[toml]>=3.4.3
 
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -27,7 +27,7 @@ classifiers =
 [options]
 packages = setuptools_rust
 zip_safe = True
-install_requires = semantic_version>=2.6.0; toml>=0.9.0
+install_requires = setuptools>=46.1; semantic_version>=2.6.0; toml>=0.9.0
```
tiran, line 30: Is 46.1 correct? I just copy and pasted the dependency from setup_requires.

By the way ``pyproject.toml`` specifies a different version.

davidhewitt: As 46.1 is the higher of the numbers, let's use that to be on the safe side.

If you're willing, could you please also update `pyproject.toml` so that they all match?

tiran: Sure!

davidhewitt (approved):
Thanks!

PR #103, opened 2021-02-08
Augment, rather than replace, existing RUSTFLAGS when building

ahesford:
Some complex environments may use RUSTFLAGS to control the behavior of rust compilation. Rather than overwriting RUSTFLAGS with custom flags, augment any existing RUSTFLAGS to preserve these environments.

This is used, for example, by Void Linux to set `--sysroot=${XBPS_CROSS_BASE}/usr` in cross-compilation environments to point to the target system root.


davidhewitt (approved):
Thanks for this fix! Mind adding a CHANGELOG entry? Then I'll merge and release this and the other fixes waiting to go out.

ahesford:
I appreciate your quick response. The latest version includes a changelog entry and a slightly more descriptive commit message.

davidhewitt:
Great, thanks again!

Issue #109, opened 2021-02-10
Shared library is missing executable bit


tiran:
``RustExtension()`` creates a shared library without executable bit. Shared library on Linux, macOS, and probably other Unix-like operation systems must have executable bit. Kernel may refuse to mmap() shared libraries as ``PROT_EXEC``.

See #106 

```
$ ldd ./build/lib.linux-x86_64-3.9/cryptography/hazmat/bindings/_rust.abi3.so
ldd: warning: you do not have execution permission for `./build/lib.linux-x86_64-3.9/cryptography/hazmat/bindings/_rust.abi3.so'
        linux-vdso.so.1 (0x00007ffce3347000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f5702f8a000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f5702f68000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f5702d9d000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f5703026000)
$ ls -la ./build/lib.linux-x86_64-3.9/cryptography/hazmat/bindings/*.so
-rwxrwxr-x. 1 heimes heimes 2669672 Feb 10 13:32 ./build/lib.linux-x86_64-3.9/cryptography/hazmat/bindings/_openssl.abi3.so
-rwxrwxr-x. 1 heimes heimes   34400 Feb 10 13:32 ./build/lib.linux-x86_64-3.9/cryptography/hazmat/bindings/_padding.abi3.so
-rw-rw-r--. 1 heimes heimes 3708904 Feb 10 13:32 ./build/lib.linux-x86_64-3.9/cryptography/hazmat/bindings/_rust.abi3.so
```

tiran:
https://github.com/PyO3/setuptools-rust/blob/991de4cc5423d4187ffd16ef185b7db10b4cb06c/setuptools_rust/build.py#L328-L332

PR #110, opened 2021-02-10
Shared libraries also need executable bit

tiran:
Shared libraries on Linux, macOS, and other Unix-like operating systems
require the executable bit, too.

Fixes: https://github.com/PyO3/setuptools-rust/issues/109


davidhewitt (approved):
👍 thanks for reporting and turning around a patch so fast. Could you please add a CHANGELOG entry for this?

davidhewitt:
👍 thanks!

PR #111, opened 2021-02-11
Don't require optional wheel dependency.

jameshilliard:
This seems to only be required for building binary wheels, so it should not be a hard requirement.

FYI I hit this issue when trying to port setuptools-rust to [buildroot](https://buildroot.org/) due to [cryptography](https://github.com/pyca/cryptography) starting to require rust, buildroot deliberately does not support binary python wheels as python packages are cross compiled from source.


davidhewitt (approved):
👍 I agree that we should probably remove the dependency. Can you please add a CHANGELOG entry to this PR? I suspect a few users will need to update their build processes once this is merged, so we should definitely notify them of this change.

jameshilliard:
> Can you please add a CHANGELOG entry to this PR?

Added to changelog.

davidhewitt:
Thanks!

Issue #108, opened 2021-02-10
"universal2" and alternate deployment target on macOS


ronaldoussoren:
Platform: macOS, in particular M1 systems.

These systems can run both arm64 (native) and x86_64 (in emulation) code, with per process selection using the "arch" command.

For C extensions it is possible to build both architectures on an M1 system (basically a form of cross compiling), including merging code for the two architectures in a fat binary ("Universal 2" in Apple-speak). 

setuptools_rust will currently always build arm64 code only, even if python was started in x86_64 mode (e.g. ``arch -x86_64 python3.9 setup.py build_ext``). 

It would be nice if setuptools_rust were able to build "universal2" wheels.

An related to this: for C extensions Python by default builds extension that work on the deployment target used for the python binary.  It might be necessary to enhance setuptools_rust to do something similar.

BTW.  I hope to have some time to work on a PR in the near future.

davidhewitt:
Thanks for reporting. If you are able to find the time, a PR to help with this would be very much appreciated.

I took a quick look just now and it looks like upstream support is possible; we'd probably need to build both targets separately and then invoke macOS sdk tools to build the final library. I'm inferring that from https://github.com/BurntSushi/ripgrep/issues/1737

davidhewitt:
Also see https://github.com/PyO3/maturin/pull/405

PR #114, opened 2021-02-14
Support building x86-64 wheel on arm64 macOS machine

messense:
This makes `arch -x86_64 python3.9 setup.py build_ext` works on Apple Silicon Macs.

cc #108 @ronaldoussoren

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -2,6 +2,7 @@ from __future__ import print_function, absolute_import
 import glob
 import json
 import os
+import platform
 import shutil
 import sys
 import subprocess
@@ -103,6 +104,9 @@ class build_rust(Command):
             target_triple = "i686-pc-windows-msvc"
         elif self.plat_name == "win-amd64":
             target_triple = "x86_64-pc-windows-msvc"
+        elif self.plat_name.startswith("macosx-") and platform.machine() == "x86_64":
+            # x86_64 or arm64 macOS targeting x86_64
+            target_triple = "x86_64-apple-darwin"
 
         if target_triple is not None:
             target_args = ["--target", target_triple]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Append to, rather than replace, existing RUSTFLAGS when building. [#103](https://github.com/PyO3/setuptools-rust/pull/103)
 - Set executable bit on shared library. [#110](https://github.com/PyO3/setuptools-rust/pull/110)
 - Don't require optional wheel dependency. [#111](https://github.com/PyO3/setuptools-rust/pull/111)
+- Support building x86-64 wheel on arm64 macOS machine. [#114](https://github.com/PyO3/setuptools-rust/pull/114)
 
 ## 0.11.6 (2020-12-13)
 


setuptools_rust/build.py
```diff
@@ -103,6 +104,10 @@ def build_extension(self, ext):
             target_triple = "i686-pc-windows-msvc"
         elif self.plat_name == "win-amd64":
             target_triple = "x86_64-pc-windows-msvc"
+        elif self.plat_name.startswith("macosx-"):
+            if platform.machine() == "arm64" and platform.processor() == "i386":
+                # arm64 macOS targeting x86_64
+                target_triple = "x86_64-apple-darwin"
```
davidhewitt, line 110: Perhaps we can allow target to be either way around? e.g. depending on the output of `platform.processor()` we could allow either  `x86_64-apple-darwin` or `aarch64-apple-darwin`

... also should it be `x86_64` instead of `i386` ?

davidhewitt (commented):
Thanks for looking into this. Gotta admit I'm not a macOS user / expert so have been reading up as best I can to give this a review...

Also please add a CHANGELOG entry 😃

setuptools_rust/build.py
messense: > Perhaps we can allow target to be either way around?

I don't think so. Only `arch -x86_64 ...` works on arm64 macOS, `arch -arm64 ...` doesn't work on x86_64 macOS.

> also should it be x86_64 instead of i386 ?

I've checked on arm64 macOS with `arch -x86_64 python3.9 setup.py build_ext` it says `i386`, but I'm not sure whether `i386` is the only possible value or not.

messense: I've also tried using [`sysctl.proc_translated`](https://developer.apple.com/documentation/apple_silicon/about_the_rosetta_translation_environment), but it doesn't seem to work for us.

x86_64 emulation, `ret` should be `1` but got `0`

native arm64, ret is `0`

```bash
❯ arch -x86_64 python -m IPython
Python 3.9.1 (default, Jan 14 2021, 04:46:20)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import platform

In [2]: platform.machine()
Out[2]: 'arm64'

In [3]: platform.processor()
Out[3]: 'i386'

In [4]: import ctypes

In [5]:
   ...: from ctypes import cdll, c_char_p, c_void_p, c_size_t, c_int, POINTER, byref
   ...:
   ...: libc = cdll.LoadLibrary("libc.dylib")
   ...: libc.sysctlbyname.argtypes = [c_char_p, c_void_p, POINTER(c_size_t), c_void_p, c_size_t]
   ...: libc.sysctlbyname.restype = c_int

In [6]: ret = c_size_t(0)

In [7]: size = c_size_t(ctypes.sizeof(ret))

In [8]: libc.sysctlbyname(b"sysctl.proc_translated", byref(ret), byref(size), None, 0)
Out[8]: 0

In [9]: ret
Out[9]: c_ulong(0)
```

```bash
❯ python -m IPython
Python 3.9.1 (default, Jan 14 2021, 04:46:20)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import ctypes

In [2]: import platform

In [3]: platform.machine()
Out[3]: 'arm64'

In [4]: platform.processor()
Out[4]: 'arm'

In [5]:
   ...: from ctypes import cdll, c_char_p, c_void_p, c_size_t, c_int, POINTER, byref
   ...:
   ...: libc = cdll.LoadLibrary("libc.dylib")
   ...: libc.sysctlbyname.argtypes = [c_char_p, c_void_p, POINTER(c_size_t), c_void_p, c_size_t]
   ...: libc.sysctlbyname.restype = c_int

In [6]: ret = c_size_t(0)

In [7]: size = c_size_t(ctypes.sizeof(ret))

In [8]: libc.sysctlbyname(b"sysctl.proc_translated", byref(ret), byref(size), None, 0)
Out[8]: 0

In [9]: ret
Out[9]: c_ulong(0)
```

messense: My bad, `sysctlbyname` should works with universal2 Python interpreter, I was testing with a arm64 only Python.

Interesting though, the arm64 only Python 3.9.1 was installed with pyenv, while `arch -x86_64 python` runs fine and results in the wired `platform.machine() == 'arm64' and platform.processor() == 'i386'`, `arch -x86_64 /absolute/path/to/python` errored with `arch: posix_spawnp: /Users/messense/.pyenv/versions/test/bin/python: Bad CPU type in executable`

With a universal2 Python interpreter, I get `platform.machine() == 'x86_64' and platform.processor() == 'i386'` when running it with `arch -x86_64 python`.

davidhewitt (approved):
Thanks, I've fixed CHANGELOG merge conflict and will merge once CI has finished.

PR #115, opened 2021-02-14
Add macOS universal2 wheel building support

### LINKED ISSUE #108: "universal2" and alternate deployment target on macOS

messense:
Based on #114 fixes #108


setuptools_rust/setuptools_ext.py
```diff
@@ -108,6 +107,14 @@ def finalize_options(self):
                         self.distribution.entry_points["console_scripts"] = ep_scripts
 
                 bdist_wheel_base_class.finalize_options(self)
+
+            def get_tag(self):
+                python, abi, plat = super().get_tag()
+                universal2 = all(ext.universal2 for ext in self.distribution.rust_extensions) \
+                    if self.distribution.rust_extensions else False
+                if universal2 and plat.startswith("macosx_"):
+                    plat = "macosx_10_9_universal2"
```
messense, line 116: I am not sure if this is the right way to set the platform tag for universal2 wheel.

README.md
```diff
@@ -133,12 +133,30 @@ It is possible to use any of the `manylinux` docker images: `manylinux1`, `manyl
 
 You can define rust extension with RustExtension class:
 
-RustExtension(name, path, args=None, features=None,
-rust\_version=None, quiet=False, debug=False)
+```python
```
davidhewitt, line 136: Thanks, this has long needed some love!

I was actually thinking about also trying to set up a readthedocs page for `setuptools-rust` - I might ask you to review it if that's ok?

README.md
```diff
@@ -147,51 +165,75 @@ The class for creating rust extensions.
...
      extension.
+   - param bool `py_limited_api`
+
+        Same as `py_limited_api` on `setuptools.Extension`. Note that if you
+        set this to True, your extension must pass the appropriate feature
+        flags to pyo3 (ensuring that `abi3` feature is enabled).
+   - param bool `universal2`
+
+        Control whether to build universal2 wheel for macOS or not.
+        Only applies to macOS targets,do nothing otherwise.
```
davidhewitt, line 236:
```suggestion
        Only applies to macOS targets,does nothing otherwise.
```

setuptools_rust/build.py
```diff
@@ -95,14 +96,18 @@ def build_extension(self, ext):
...
+            if os.getenv("CARGO_BUILD_TARGET"):
+                target_triple = os.environ["CARGO_BUILD_TARGET"]
+            elif self.plat_name == "win32":
+                target_triple = "i686-pc-windows-msvc"
+            elif self.plat_name == "win-amd64":
+                target_triple = "x86_64-pc-windows-msvc"
+            elif self.plat_name.startswith("macosx-"):
+                if platform.machine() == "arm64" and platform.processor() == "i386":
+                    # arm64 macOS targeting x86_64
+                    target_triple = "x86_64-apple-darwin"
```
davidhewitt, line 110: Should this logic maybe be moved into a `get_target_triple()` function? It's getting quite long.

setuptools_rust/setuptools_ext.py
davidhewitt: It looks like `sysconfig.get_platform()` might return something with `universal2` in it, which could be better? https://github.com/pypa/setuptools/issues/2520

davidhewitt (commented):
It looks like `sysconfig.get_platform()` might return something with `universal2` in it on some macOS versions: pypa/setuptools#2520

... rather than adding a `universal2` option, should we be detecting this and building `universal2` wheels automatically? 

cc @ronaldoussoren in case you know about this?

README.md
messense: Sure thing!

setuptools_rust/setuptools_ext.py
messense: That only returns `macosx-10.9-universal2` on macOS universal2 Python build, does not play well with cross compilation.

messense: It doesn't return `universal2` with Apple's bundled Python in Xcode which is universal2 build itself.

```bash
❯ python3
Python 3.8.2 (default, Dec 21 2020, 15:06:03)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
>>> import sys
>>> sys.prefix
'/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8'
>>> import sysconfig; sysconfig.get_platform()
'macosx-10.14.6-arm64'

❯ file /Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/bin/python3.8
/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/bin/python3.8: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/bin/python3.8 (for architecture x86_64):	Mach-O 64-bit executable x86_64
/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework/Versions/3.8/bin/python3.8 (for architecture arm64):	Mach-O 64-bit executable arm64
```

ronaldoussoren: > It looks like `sysconfig.get_platform()` might return something with `universal2` in it, which could be better? [pypa/setuptools#2520](https://github.com/pypa/setuptools/issues/2520)

Sysconfig.get_platform() will return "universal2" as the "CPU architecture" when Python was build as a universal 2 binary using the configure script.  This is true for the (currently experimental) universal2 installer for Python 3.9.1 on Python.org, as wel as recent 3.10 alpha installers. 

The python included in Xcode does not return "universal2" in the platform tag, likely because it was build twice and lipo-d together.

My reason for filing #108 is to have a way to build universal2 wheels that match the python.org installers (including a deployment target of macOS 10.9 if that is at all possible using the default rust toolchain). Such wheels would also be compatible with other macOS builds of CPython (such as the ones in Xcode and home-brew). 

Having support for this in this project is a prerequisite for me asking users of setuptools-rust to provide universal2 binary wheels (although there's also the question of CI support, none of the large cloud CI systems currently have Apple M1 runners)

messense: I think to set deployment target to macOS 10.9 you just need to set environment variable `MACOSX_DEPLOYMENT_TARGET=10.9`. https://stackoverflow.com/a/64864364

messense:
Rebased and added tests.

davidhewitt (commented):
Thanks, and sorry for the slow review again by me.

What do you think if instead of setting `universal2=True` in the `RustExtension` class, we use an environment variable (or an option to `bdist_wheel`) to configure this?

This is because I think that users running `python setup.py install` won't care at all about whether their binary is `universal2` and might just find it harder to build if they start getting error messages about `lipo` and `fat-macho`.

Really the only time this setting is relevant (I think?) is when package maintainers are running their CI jobs to build wheels.

It looks like `ARCHFLAGS` might be an appropriate environment variable to look at: https://stackoverflow.com/questions/64364310/how-do-i-resolve-error-architecture-not-supported-during-pip-install-psutil

.github/workflows/ci.yml
```diff
@@ -104,6 +112,18 @@ jobs:
           python setup.py bdist_wheel --py-limited-api=cp35
           ls -la dist/
 
+      - name: Build an universal2 abi3 wheel
+        if: matrix.os == 'macos-latest'
+        env:
+          MACOSX_DEPLOYMENT_TARGET: '10.9'
```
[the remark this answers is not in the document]
messense: I think user should control this explicitly.

messense:
I was aware of `ARCHFLAGS`, changed to use that. Now `setuptools-rust` should works well with `cibuildwheel` since they also use `ARCHFLAGS` to build universal2 wheel.

davidhewitt (approved):
Thanks for the multiple revisions to this PR, this looks great to me now!

I think it needs some documentation, but there's not really a good place to document it so I'll do that when I write some readthedocs documentation soon.

PR #117, opened 2021-02-15
doc: improve error message for missing Rust compiler

davidhewitt:
This PR tries to help out with a more helpful error message for users who don't have a working Rust compiler, and may even not know they need one.

cc @alex  - I think this tries to capture the common cases you've experienced in cryptography. I'd appreciate any feedback you have on whether you think this error message is helpful, too long, suggested wording changes, or any other thoughts you have on it.

Example output a cryptography user might see:

```
error: can't find Rust compiler

If you are using an outdated pip version or do not have `wheel` installed, it is possible a prebuilt wheel is available for
this package but pip is not able to install from it. Installing from the wheel would avoid the need for a Rust compiler.

To update pip and wheel, run:

    pip install --upgrade pip wheel

and then retry package installation.

If you did intend to build this package from source, try installing a Rust compiler from your system package manager and ens
ure it is on the PATH during installation. Alternatively, rustup (available at https://rustup.rs) is the recommended way to
download and update the Rust compiler toolchain.

This package requires Rust >=1.41.0.
```

### DIFF
--- a/setuptools_rust/utils.py
+++ b/setuptools_rust/utils.py
@@ -57,12 +57,32 @@ def rust_features(ext=True, binding=Binding.PyO3):
         raise DistutilsPlatformError(f"unknown Rust binding: '{binding}'")
 
 
-def get_rust_version():
+def get_rust_version(min_version=None):
     try:
         output = subprocess.check_output(["rustc", "-V"]).decode("latin-1")
         return semantic_version.Version(output.split(" ")[1], partial=True)
     except (subprocess.CalledProcessError, OSError):
-        raise DistutilsPlatformError("can't find Rust compiler")
+        raise DistutilsPlatformError(
+            "can't find Rust compiler\n\n"
+            "If you are using an outdated pip version, it is possible a "
+            "prebuilt wheel is available for this package but pip is not able "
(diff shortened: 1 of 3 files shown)


setuptools_rust/utils.py
```diff
@@ -57,12 +57,32 @@ def rust_features(ext=True, binding=Binding.PyO3):
...
     except (subprocess.CalledProcessError, OSError):
-        raise DistutilsPlatformError("can't find Rust compiler")
+        raise DistutilsPlatformError(
+            "can't find Rust compiler\n\n"
+            "If you are using an outdated pip version or do not have `wheel` "
+            "installed, it is possible a prebuilt wheel is available for this "
+            "package but pip is not able to install from it. Installing from "
+            "the wheel would avoid the need for a Rust compiler.\n\n"
+            "To update pip and wheel, run:\n\n"
+            "    pip install --upgrade pip wheel\n\n"
```
alex, line 72: Wheel isn't required on the installer side, pip contains everything required to install wheels. Besides dropping this, I think this is great!

davidhewitt: 👍 thanks, I think maybe in the past it was, at some point? I got confused by some documentation online...

PR #119, opened 2021-02-18
Set a more reasonable LC_ID_DYLIB entry on macOS

### LINKED ISSUE #106: Rust binding in Cryptography 3.4+ dynamically links with non-existent library on macOS
Issue cloned from the Cryptography issue tracker per request, as setuptools-rust handles this for them: https://github.com/pyca/cryptography/issues/5807

The Cryptography wheel for macOS includes _rust.abi3.so. This references a library that isn't shipped and has a path clearly from the build machine:
```
(foo) dpage@hal:~/git/pgadmin4/foo/lib/python3.8/site-packages/cryptography/hazmat/bindings$ otool -L _rust.abi3.so 
_rust.abi3.so:
	/Users/runner/work/cryptography/cryptography/cryptography-3.4.4/src/rust/target/release/deps/libcryptography_rust.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/usr/lib/libresolv.9.dylib (compatibility version 1.0.0, current version 1.0.0)
```

Whilst this may not break the package, it:

A) Is not good practice
B) Breaks build systems that automatically fixup library paths when creating relocatable appbundles.

B is the issue I'd like to see fixed, as the current situation requires me to modify code that's worked for many years (15+) to ignore errors when processing this particular file.

An additional issue, again likely benign for the general use case, is that _rust.abi3.so does not have the executable bit set as is generally the norm for shared libraries on *nix platforms. This also required me to change build code, as searches for binaries that need to be codesigned filtered possible targets based on the executable bit, before doing further tests to see if the file was actually an executable or library (which gets *very* expensive without that check, in a codebase with thousands of Python/JS/HTML files etc).

Probably not relevant, but here's my system info:
```
(foo) dpage@hal:~/git/pgadmin4/foo/lib/python3.8/site-packages/cryptography/hazmat/bindings$ pip freeze
cffi==1.14.4
cryptography==3.4.4
pycparser==2.20
(foo) dpage@hal:~/git/pgadmin4/foo/lib/python3.8/site-packages/cryptography/hazmat/bindings$ python --version
Python 3.8.2
(foo) dpage@hal:~/git/pgadmin4/foo/lib/python3.8/site-packages/cryptography/hazmat/bindings$ pip --version
pip 21.0.1 from /Users/dpage/git/pgadmin4/foo/lib/python3.8/site-packages/pip (python 3.8)
(foo) dpage@hal:~/git/pgadmin4/foo/lib/python3.8/site-packages/cryptography/hazmat/bindings$ sw_vers -productVersion
11.1
```

Thanks!

messense:
Fixes #106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### Fixed
 - Set executable bit on shared library. [#110](https://github.com/PyO3/setuptools-rust/pull/110)
 - Don't require optional `wheel` dependency. [#111](https://github.com/PyO3/setuptools-rust/pull/111)
+- Set a more reasonable LC_ID_DYLIB entry on macOS. [#119](https://github.com/PyO3/setuptools-rust/pull/119)
 
 ## 0.11.6 (2020-12-13)
 
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -180,6 +180,10 @@ class build_rust(RustCommand):
         if ext.native:
             rustflags += " -C target-cpu=native"
 
+        if not executable and sys.platform == "darwin":
+            ext_basename = os.path.basename(self.get_dylib_ext_path(ext, ext.name))
+            rustflags += f" -C link-args=-Wl,-install_name,@rpath/{ext_basename}"
+
         if rustflags:
             env["RUSTFLAGS"] = (env.get("RUSTFLAGS", "") + " " + rustflags).strip()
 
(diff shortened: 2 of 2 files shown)


davidhewitt (approved):
Looks great to me, thanks very much for this!

messense:
@konstin We might want to do the same thing for maturin.

konstin:
@messense Sounds good, unfortunately I don't have a mac so I can't myself

mathstuf:
Could this be released so that wheels for `cryptography` with usable library IDs then be released? Alternatively, is there already a schedule in place I could track?

davidhewitt:
@mathstuf it'll be coming very soon - see https://github.com/PyO3/setuptools-rust/issues/106#issuecomment-791010981

PR #121, opened 2021-03-07
docs: add sphinx / readthedocs configuration

davidhewitt:
This PR adds a first pass at sphinx documentation for the project. There's definitely more that could be added, however I think this makes a nice start before we release version 0.12 . 

[Rendered](https://setuptools-rust.readthedocs.io/en/readthedocs/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,13 @@
 # Setuptools plugin for Rust extensions
 
-[![Build Status](https://travis-ci.org/PyO3/setuptools-rust.svg?branch=master)](https://travis-ci.org/PyO3/setuptools-rust)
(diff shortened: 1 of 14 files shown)


messense (approved):
Looks great!

README.md
```diff
@@ -129,70 +133,6 @@ You can then upload the `manylinux2014` wheels to pypi using [twine](https://git
 
 It is possible to use any of the `manylinux` docker images: `manylinux1`, `manylinux2010` or `manylinux2014`. (Just replace `manylinux2014` in the above instructions with the alternative version you wish to use.)
```
messense, line 135: I think we should add a section about how to build macOS universal2 binary wheels here.

davidhewitt: Agreed, I'll add it later today in a follow-up PR 👍

PR #134, opened 2021-03-20
Add support for cross

### LINKED ISSUE #10: Support "cross"
[cross](https://github.com/japaric/cross) is a great tool for cross compile Rust code to other platform, support it will be awesome ! (so build Python wheel for more platform will be easier)

modification we may need:

* let user can build wheel for different target platform (need to pass `--target $TARGET` to rustc)
* install `cross`, this will be required only if you want to use `cross` feature
* let build command pass `cross` instead of `cargo` (if user choose `cross`)

messense:
Closes #10

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -50,6 +50,7 @@ class build_rust(RustCommand):
         self.build_temp = None
         self.plat_name = None
         self.target = os.getenv("CARGO_BUILD_TARGET")
+        self.cargo = os.getenv("CARGO", "cargo")
 
     def finalize_options(self):
         super().finalize_options()
@@ -123,7 +124,7 @@ class build_rust(RustCommand):
 
         # Find where to put the temporary build files created by `cargo`
         metadata_command = [
-            "cargo",
+            self.cargo,
             "metadata",
             "--manifest-path",
             ext.path,
@@ -153,7 +154,7 @@ class build_rust(RustCommand):
 
         if executable:
             args = (
-                ["cargo", "build", "--manifest-path", ext.path]
+                [self.cargo, "build", "--manifest-path", ext.path]
                 + feature_args
                 + target_args
                 + list(ext.args or [])
@@ -169,7 +170,7 @@ class build_rust(RustCommand):
 
         else:
             args = (
-                ["cargo", "rustc", "--lib", "--manifest-path", ext.path]
+                [self.cargo, "rustc", "--lib", "--manifest-path", ext.path]
                 + feature_args
                 + target_args
                 + list(ext.args or [])
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -124,7 +124,7 @@ class build_rust(RustCommand):
 
         # Find where to put the temporary build files created by `cargo`
         metadata_command = [
-            self.cargo,
+            "cargo",
             "metadata",
             "--manifest-path",
             ext.path,
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -4,6 +4,7 @@ import os
 import platform
 import shutil
 import sys
+import sysconfig
 import subprocess
 from distutils.errors import (
     CompileError,
@@ -350,9 +351,41 @@ class build_rust(RustCommand):
         assert modpath not in build_ext.ext_map
         build_ext.ext_map[modpath] = ext
         try:
-            return build_ext.get_ext_fullpath(target_fname)
+            ext_path = build_ext.get_ext_fullpath(target_fname)
         finally:
             del build_ext.ext_map[modpath]
+
+        if '.abi3.' in ext_path:
+            return ext_path
+        # Examples: linux_x86_64, linux_i686, manylinux2014_aarch64, manylinux_2_24_armv7l
+        plat_name = self.plat_name.lower().replace('-', '_').replace('.', '_')
+        if not plat_name.startswith(('linux', 'manylinux')):
+            return ext_path
+
+        arch_parts = []
+        arch_found = False
+        for item in plat_name.split('_'):
+            if item.startswith(('linux', 'manylinux')):
+                continue
+            if item.isdigit() and not arch_found:
+                # manylinux_2_24_armv7l arch should be armv7l
+                continue
+            arch_found = True
+            arch_parts.append(item)
+        target_arch = '_'.join(arch_parts)
+        host_platform = sysconfig.get_platform()
+        host_arch = host_platform.rsplit('-', 1)[1]
+        # Remove incorrect platform tag if we are cross compiling
+        if target_arch and host_arch != target_arch:
+            ext_dir = os.path.dirname(ext_path)
+            ext_name = os.path.basename(ext_path)
+            # rust.cpython-38-x86_64-linux-gnu.so to (rust.cpython-38-x86_64-linux-gnu, .so)
+            ext_name, ext_ext = os.path.splitext(ext_name)
+            # rust.cpython-38-x86_64-linux-gnu to (rust, .cpython-38-x86_64-linux-gnu)
+            ext_name, _plat_tag = os.path.splitext(ext_name)
+            # rust.so, removed platform tag
+            ext_path = os.path.join(ext_dir, ext_name + ext_ext)
+        return ext_path
 
     @staticmethod
     def create_universal2_binary(output_path, input_paths):
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,3 +185,18 @@ jobs:
           python -m pip install wheel
           python setup.py bdist_wheel --plat-name manylinux2014_aarch64
           ls -la dist/
+      - uses: uraimo/run-on-arch-action@v2.0.5
+        name: Install built wheel
+        with:
+          arch: aarch64
+          distro: ubuntu20.04
+          dockerRunArgs: |
+            --volume "${PWD}/examples/namespace_package:/io"
+          install: |
+            apt-get update
+            apt-get install -y --no-install-recommends python3 python3-pip
+            pip3 install -U pip
+          run: |
+            pip3 install namespace_package --no-index --find-links /io/dist/ --force-reinstall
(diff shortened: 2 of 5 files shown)


messense:
```
creating build/bdist.linux-x86_64
creating build/bdist.linux-x86_64/wheel
creating build/bdist.linux-x86_64/wheel/namespace_package
creating build/bdist.linux-x86_64/wheel/namespace_package/python
copying build/lib/namespace_package/python/__init__.py -> build/bdist.linux-x86_64/wheel/namespace_package/python
copying build/lib/namespace_package/rust.cpython-38-x86_64-linux-gnu.so -> build/bdist.linux-x86_64/wheel/namespace_package
running install_egg_info
running egg_info
creating namespace_package.egg-info
writing namespace_package.egg-info/PKG-INFO
writing dependency_links to namespace_package.egg-info/dependency_links.txt
writing top-level names to namespace_package.egg-info/top_level.txt
writing manifest file 'namespace_package.egg-info/SOURCES.txt'
reading manifest file 'namespace_package.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'namespace_package.egg-info/SOURCES.txt'
Copying namespace_package.egg-info to build/bdist.linux-x86_64/wheel/namespace_package-0.1.0-py3.8.egg-info
running install_scripts
creating build/bdist.linux-x86_64/wheel/namespace_package-0.1.0.dist-info/WHEEL
creating 'dist/namespace_package-0.1.0-cp38-cp38-manylinux2014_aarch64.whl' and adding 'build/bdist.linux-x86_64/wheel' to it
adding 'namespace_package/rust.cpython-38-x86_64-linux-gnu.so'
```

Looks like https://github.com/PyO3/setuptools-rust/issues/138#issuecomment-807964452

setuptools_rust/build.py
```diff
@@ -349,10 +351,42 @@ def get_dylib_ext_path(self, ext, target_fname):
...
+        host_platform = sysconfig.get_platform()
+        host_arch = host_platform.rsplit('-', 1)[1]
+        # Remove incorrect platform tag if we are cross compiling
+        if target_arch and host_arch != target_arch:
+            ext_dir = os.path.dirname(ext_path)
+            ext_name = os.path.basename(ext_path)
+            # rust.cpython-38-x86_64-linux-gnu.so to (rust.cpython-38-x86_64-linux-gnu, .so)
+            ext_name, ext_ext = os.path.splitext(ext_name)
+            # rust.cpython-38-x86_64-linux-gnu to (rust, .cpython-38-x86_64-linux-gnu)
+            ext_name, _plat_tag = os.path.splitext(ext_name)
```
davidhewitt, line 385: We do a similar pattern about line ~310, maybe can refactor to share a common util function?

setuptools_rust/build.py
```diff
@@ -349,10 +351,42 @@ def get_dylib_ext_path(self, ext, target_fname):
...
+            if item.isdigit() and not arch_found:
+                # manylinux_2_24_armv7l arch should be armv7l
+                continue
+            arch_found = True
+            arch_parts.append(item)
+        target_arch = '_'.join(arch_parts)
+        host_platform = sysconfig.get_platform()
+        host_arch = host_platform.rsplit('-', 1)[1]
+        # Remove incorrect platform tag if we are cross compiling
+        if target_arch and host_arch != target_arch:
```
davidhewitt, line 379: Yikes. Do we also need to check abi here? e.g. `windows-gnu` versus `windows-msvc`?

Is it enough to check if `plat_name` doesn't match `sysconfig.get_platform()`?

Maybe we should remove the platform tag always, rather than try to do some clever detection? Depends if there's a reason to keep the platform tag around.

davidhewitt (commented):
Thanks for implementing this, and also the great example / test.

I've had to do a lot of reading to get my head around all of this. I think I understand now, and I'll summarise below. Please correct any of this which is wrong:
1. `cross build` will run the build command inside a dockerfile which contains `rustc` configured to cross-compile.
   - The dockerfile is still running the host architecture.
2. In the example, we put a python installation inside the dockerfile so we can point `PYO3_CROSS_LIB_DIR` at it.
   - I think here we get lucky: we're targeting ARM but we're pointing PyO3 to a python installation on the host architecture.
3. We have to remove the platform tag, because Python isn't aware we're cross compiling and so gives us the wrong extension filename.
4. We use a github action which emulates ARM architecture to test the built wheel works.

So I have a concern about (2) that we're lucky we're getting a working build; PyO3 could easily get some wrong information from the `PYO3_CROSS_LIB_DIR` we provide it - it's for the host architecture, not the target architecture.

Maybe it's better for the example to cross-compile for a `--py-limited-api` where we don't need a cross lib dir? We might not need a custom Dockerfile at all in that case.

Alternatively I wonder if we can combine it with #139 to somehow set up `crossenv` inside the Dockerfile?

setuptools_rust/build.py
davidhewitt: Or alternatively, looking at discussion in #138 it sounds like Python also gets this wrong? Maybe we should not fix the platform tag, and instead just leave it to users. If that's consistent with what C does, it seems most reliable (even if it's annoying).

messense:
>2.  In the example, we put a python installation inside the dockerfile so we can point PYO3_CROSS_LIB_DIR at it.
> I think here we get lucky: we're targeting ARM but we're pointing PyO3 to a python installation on the host architecture.

We didn't get lucky. The `PYO3_CROSS_LIB_DIR` is passed into Docker container where it is point to an aarch64 Python build from `quay.io/pypa/manylinux2014_aarch64` docker image which is really aarch64 .

setuptools_rust/build.py
messense: > Maybe we should remove the platform tag always, rather than try to do some clever detection?

I don't know, but I did encounter an error once with PyPy with no platform tag, see https://github.com/PyO3/setuptools-rust/runs/2208718989?check_suite_focus=true

messense:
> Maybe it's better for the example to cross-compile for a --py-limited-api where we don't need a cross lib dir? We might not need a custom Dockerfile at all in that case.

We can certainly do that. But I think it's useful to have an example without `abi3`.

> Alternatively I wonder if we can combine it with #139 to somehow set up crossenv inside the Dockerfile?

crossenv currently has [an issue with Rust built scripts](https://github.com/benfogle/crossenv/issues/65) which is partially addressed in https://github.com/benfogle/crossenv/pull/66

setuptools_rust/build.py
messense: > Or alternatively, looking at discussion in #138 it sounds like Python also gets this wrong? 

Yes, it's terrible.

> Maybe we should not fix the platform tag, and instead just leave it to users. If that's consistent with what C does, it seems most reliable (even if it's annoying).

It would make supporting `cross` kinda useless IMO.

messense:
BTW, using crossenv along with cross isn't really useful, we would also need a copy of aarch64 Python outside of Docker container to create the cross compilation virtualenv because we invoke `python setup.py bdist_wheel` outside of Docker container.

messense:
Note that I'm not keen on pursue this as cross official Docker images are not manylinux complaint.

davidhewitt:
> aarch64 Python build from quay.io/pypa/manylinux2014_aarch64 docker image 

Ah, yes I missed that the manylinux image chosen was for arm 👍 

> BTW, using crossenv along with cross isn't really useful, we would also need a copy of aarch64 Python outside of Docker container to create the cross compilation virtualenv because we invoke python setup.py bdist_wheel outside of Docker container.

Yeah that makes sense, thanks.

> Note that I'm not keen on pursue this as cross official Docker images are not manylinux complaint.

You mean that you think we should close this PR instead of merging? I wonder how hard it would be for us to maintain a set of Docker images for cross. But maybe it's easier for us to instead work on buildroot + crossenv support?

messense:
> You mean that you think we should close this PR instead of merging?

I'm fine with closing this PR. But it could be useful to build statically linked musl-libc executables which is manylinux compliant.

>  I wonder how hard it would be for us to maintain a set of Docker images for cross.

I'm not sure, cross relies on a lot of things in its Docker container. https://github.com/rust-embedded/cross/tree/master/docker

> But maybe it's easier for us to instead work on buildroot + crossenv support?

I think so.

wdv4758h:
I just see someone tagged me, didn't read through whole threads.

Good to see someone is still pushing this. I think Python's extension cross compile story is bad for quite long time ...
I don't do extension cross compile recently, but I can see some tricks I used couple years ago like removing wrong platform tag (was doing cross compile for extension in C).

setuptools_rust/build.py
davidhewitt: Still would like to see this, maybe call `strip_platform_and_extension()` ?

setuptools_rust/build.py
```diff
@@ -349,10 +351,42 @@ def get_dylib_ext_path(self, ext, target_fname):
         assert modpath not in build_ext.ext_map
         build_ext.ext_map[modpath] = ext
         try:
-            return build_ext.get_ext_fullpath(target_fname)
+            ext_path = build_ext.get_ext_fullpath(target_fname)
         finally:
             del build_ext.ext_map[modpath]
 
+        if '.abi3.' in ext_path:
```
davidhewitt, line 358: The logic from here down is quite complex. Can we split it into a few smaller functions, to help with cognitive load? e.g. `is_linux_cross_compile()` , `strip_platform_and_extension()`? That might be enough.

davidhewitt (commented):
Sorry I delayed this for so long.

@messense has your opinion changed? I see you now maintain https://github.com/messense/manylinux-cross, so perhaps with good documentation we can help users use `cross` correctly?

If so, let's tidy this up and merge it :)

messense:
It's certainly useful for bin bindings, feel free to take over and get it to finish line, thanks!

morenol:
I just tried this branch and it works for me to cross-compile a python package for a raspberry pi. Is there something that I can do to help this land to the main branch?

messense:
> I just tried this branch and it works for me to cross-compile a python package for a raspberry pi. Is there something that I can do to help this land to the main branch?

A rebased version of this PR maybe?

PR #135, opened 2021-03-20
Build README has the wrong pip arguments (minor text fix)

corydodt:
Just a little error in the README, no `-r` flag if you're not using a requirements.txt file.


davidhewitt:
Nice catch, thanks for fixing!

PR #136, opened 2021-03-21
Add --target command line option for specify target triple

messense:
This mimics the same `--target` option of cargo and maturin, it's much more discoverable than the `CARGO_BUILD_TARGET` environment variable IMO.


davidhewitt:
Thanks very much for this; agreed it's a nice option to have.

Issue #132, opened 2021-03-20
1.0 release?


davidhewitt:
I think that `setuptools-rust` works well as-is, so it could be a good time to consider a 1.0 release.

Some things that I propose checking / tweaking before releasing as 1.0:

- [x] `py_limited_api` argument to `RustExtension` should perhaps be removed and instead we pick up the command line flag already supported by setuptools e.g. `python setup.py bdist_wheel --py-limited-api=cp36` - and we would pass the `abi3-py36` flag to `pyo3` automatically
- [x] move wheel-building documentation from the README into a separate page of the sphinx documentation. (There are a few possibilities for wheel building, such as `build-wheels.sh` and `cibuildwheel`, so it would be nice to expand on these.)

davidhewitt:
cc @messense what do you think about 1.0 ?

cc @alex would the proposed change to the `py_limited_api` be fine for your usage in `cryptography`? Reusing the existing command line flag seems desirable.

messense:
Better cross compiling support code-wise or documentation-wise, possibly direct people to use [crossenv](https://github.com/benfogle/crossenv).

alex:
When I added `py_limited_api`, I modeled it on how it works for C extensions, which is namely that you have three separate things:

- naming the `.whl`
- naming the `.so`
- and defining `PY_LIMITED_API` in C

Doing all these separately seems error prone, but I guess I don't have a strong opinion on the value of being consistent with setuptools vs. improving that UX.

davidhewitt:
#163

PR #137, opened 2021-03-21
Pass pyo3/abi3-pyXX feature to pyo3 automatically

messense:


davidhewitt:
Thank you for all of these PRs! I have some time later this afternoon so will review all of them then.

setuptools_rust/build.py
```diff
@@ -138,6 +138,11 @@ def build_extension(self, ext: RustExtension, target_triple=None):
 
         features = set(ext.features)
         features.update(rust_features(binding=ext.binding))
+        if ext.py_limited_api and ext.binding == Binding.PyO3 and platform.python_implementation() != 'PyPy':
```
davidhewitt, line 141: I don't think we need to check about PyPy here. PyO3 will warn if using PyPy is inappropriate, and PyPy will support in a future version.

setuptools_rust/setuptools_ext.py
```diff
@@ -162,6 +162,7 @@ class bdist_wheel_rust_extension(bdist_wheel_base_class):
             def finalize_options(self):
                 scripts = []
                 for ext in self.distribution.rust_extensions:
+                    ext.py_limited_api = self.py_limited_api
```
davidhewitt, line 165: Hmm I'd rather we didn't modify the `RustExtension` objects. I think we can do this an alternative way by loading the `bdist_wheel` command when we want to decide whether to build a limited api wheel.

davidhewitt (commented):
Thanks! I took a long time to think about this, and decided to create a PR to add some extra functionality on top of this commit. See https://github.com/messense/setuptools-rust/pull/1

I made it a separate PR onto your branch because I wanted to give you a chance to review it.

Once that is merged into this branch, I'm happy to push a CHANGELOG entry and more docs here.

setuptools_rust/build.py
davidhewitt: (fixed in my PR)

setuptools_rust/setuptools_ext.py
davidhewitt: (fixed in my PR)

examples/rust_with_cffi/setup.py
```diff
@@ -19,11 +19,7 @@
     ],
     packages=["rust_with_cffi"],
     rust_extensions=[
-        RustExtension(
-            "rust_with_cffi.rust",
-            py_limited_api=True,
-            features=[] if platform.python_implementation() == 'PyPy' else ["pyo3/abi3"]
-        ),
+        RustExtension("rust_with_cffi.rust"),
```
messense, line 22: I think we should use auto here.

davidhewitt: 👍 sounds reasonable. `"auto"` is the default, but can't hurt for us to be explicit to help users discover it.

davidhewitt:
@alex based off https://github.com/PyO3/setuptools-rust/issues/132#issuecomment-803356036, I'd be interested in hearing what you think of this PR.

Essentially if you set `py_limited_api` to `True` or `False`, the behaviour will be backwards-compatible and continue to match setuptools - you'll need to define the features as currently done.

With the new default "auto" mode, the command-line `--py-limited-api` option to `bdist_wheel` will be used to infer the build mode for the extension too, and will pass the correct feature to pyo3. I'd be curious if you'd want to use this in `cryptography`.

IMO it seems like a win for everyone: package maintainers can still build and distribute abi3 wheels, and users installing the package directly can build a version-specific extension. (PyO3 already uses a few optimizations when building version-specific, and I'd consider adding more.)

alex:
I think we'd probably use this in pyca/cryptography, sure.

One note: I think you'll want to have auto -> not try to enable abi3 on PyPy still.

davidhewitt:
Great!

> One note: I think you'll want to have auto -> not try to enable abi3 on PyPy still.

Given that PyPy is sounding like it'll support a "stable api" for PyPy 3.8, my thought was that we should allow this in `setuptools-rust` so that it's future-compatible with when PyPy does add support.

PyO3 0.14 will emit a warning when abi3 is enabled for PyPy; we opted to do that so that it wouldn't cause a hard error in builds if the feature is set. PyO3 0.13.2 will silently complete the build, which imo is still fine because it'll be a working build. Setuptools just won't actually package it as an abi3 wheel.

PR #141, opened 2021-03-29
Add cross compilation test with crossenv

messense:


messense:
Not sure why cross compiling to aarch64 works but fails with armv7. It seems that `pip install cffi` did not use manylinux1 wheel: https://github.com/PyO3/setuptools-rust/runs/2217445352?check_suite_focus=true

@benfogle do you know why?

davidhewitt:
> It seems that pip install cffi did not use manylinux1 wheel: 

I can't see a `linux-arm` wheel here at https://pypi.org/project/cffi/#files - presumably that is the cause? There _is_ a wheel there for aarch64.

From some testing inside your docker image manually I was able to get cffi to install in the cross-env with the following command:

```
CPPFLAGS="-I /usr/armv7-unknown-linux-gnueabihf/include/ -L /usr/armv7-unknown-linux-gnueabihf/lib/" pip install cffi
```

Perhaps these flags need to be part of the cross-compilation's `./configure` call, or maybe passed when making the cross-env?

benfogle:
Well, crossenv 1.1.0 disabled all manylinux wheels by default, because the detection algorithms are broken. I just added a `--manylinux` option back in for 1.1.1 to specifically allow them. Unfortunately, it still won't work for armv7, because of [this](https://github.com/pypa/packaging/blob/b5878c977206f60302536db969a8cef420853ade/packaging/tags.py#L646). It tries to verify that it's a compatible arm version by reading ELF headers from `sys.executable`, which is never going to work.

For this, you can try:
```
build-pip install cffi
cross-expose cffi  # tell cross-pip that cffi is installed
...
```

messense:
> I can't see a linux-arm wheel here at https://pypi.org/project/cffi/#files - presumably that is the cause? There is a wheel there for aarch64.

It's only needed in `setup.py` when running `bdist_wheel` (It's in `setup_requires`), so it's fine to use a x86_64 version I think, very much like Rust build scripts.

messense:
> Well, crossenv 1.1.0 disabled all manylinux wheels by default, because the detection algorithms are broken. I just added a `--manylinux` option back in for 1.1.1 to specifically allow them. Unfortunately, it still won't work for armv7, because of [this](https://github.com/pypa/packaging/blob/b5878c977206f60302536db969a8cef420853ade/packaging/tags.py#L646). It tries to verify that it's a compatible arm version by reading ELF headers from `sys.executable`, which is never going to work.
> 
> For this, you can try:
> 
> ```
> build-pip install cffi
> cross-expose cffi  # tell cross-pip that cffi is installed
> ...
> ```

It works, thanks!

.github/workflows/ci.yml
```diff
@@ -155,3 +155,49 @@ jobs:
...
+      - name: Build Wheels
+        run: |
+          echo 'curl -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain stable
+          source ~/.cargo/env
+          rustup target add ${{ matrix.platform.target }}
+          cd examples/rust_with_cffi/
+          python3.6 -m pip install "crossenv>=1.1.1"
+          python3.6 -m crossenv "/opt/python/cp36-cp36m/bin/python3" --cc $TARGET_CC --cxx $TARGET_CXX --sysroot $TARGET_SYSROOT --env LIBRARY_PATH= --manylinux manylinux1 venv
+          . venv/bin/activate
+          export PYO3_CROSS_LIB_DIR="$PWD/venv/lib"
```
messense, line 178: I was expecting this to be set by
https://github.com/PyO3/setuptools-rust/blob/503c3390312a904850d3711e5ca9dc685ac2221a/setuptools_rust/build.py#L178

but seems that it didn't.

messense:
```python
(cross) root@506788c08417:/io/examples/rust_with_cffi# python
Python 3.6.12 (default, Mar 29 2021, 06:48:42)
[GCC 9.3.0] on linux
>>> import sysconfig
>>> sysconfig.get_config_var('HOST_GNU_TYPE')
'aarch64-unknown-linux-gnu'
>>> sysconfig.get_config_var('BUILD_GNU_TYPE')
'aarch64-unknown-linux-gnu'
>>> sysconfig.get_path('stdlib')
'/opt/python/cp36-cp36m/lib/python3.6'
>>>
```

benfogle: Huh. It looks like host-python was natively compiled on an aarch64 system. That is something I can fix on my end.

messense: Yes, it was copied from manylinux2014_aarch64 docker image: https://github.com/messense/manylinux2014-cross-arm/blob/1da06861a677d9e65e38b51557080e3c505245ff/aarch64.Dockerfile#L133-L134

benfogle: That same issue on my end is also responsible for you needing to specify `--plat-name` to `bdist_wheel` as well.

messense: Found another issue, since I've set `CARGO_BUILD_TARGET` it's not going to use `get_nix_target_info()`. 😅

https://github.com/PyO3/setuptools-rust/blob/503c3390312a904850d3711e5ca9dc685ac2221a/setuptools_rust/build.py#L80-L81

messense: > That same issue on my end is also responsible for you needing to specify `--plat-name` to `bdist_wheel` as well.

The platform name seems is wrong for armv7l, https://github.com/PyO3/setuptools-rust/pull/141/checks?check_run_id=2242561569

```
ERROR: rust_with_cffi-0.1.0-cp36-abi3-linux_arm.whl is not a supported wheel on this platform.
```

benfogle: I believe that's an artifact of how that version of host-python was compiled. I checked in the docker image, and it was built with `HOST_GNU_TYPE` of `armv7-unknown-linux-gnueabihf`.  The string `armv7l` doesn't appear _anywhere_ in either the compiled binaries or Python source. On a real system, it would get the armv7l from the running operating system.

In this case it seems like the easiest solution would be to pass `--plat-name` as you've been doing, or convince the emulated system your installing it on to accept those wheels. I can also try to add a `--machine` option to crossenv, but it doesn't seem like that actually makes things easier for you: It just changes where you have to add options.

messense: I agree. That's why I've pushed https://github.com/messense/manylinux2014-cross-arm/commit/c0e73ddb5e8a4d04f76573993fe095601fee1a2a to force `_PYTHON_HOST_PLATFORM=linux-armv7l`

setuptools_rust/build.py
```diff
@@ -77,9 +79,7 @@ def get_target_info(self):
         # we'll target a 32-bit Rust build.
         # Automatic target detection can be overridden via the CARGO_BUILD_TARGET
         # environment variable or --target command line option
-        if self.target:
-            return _TargetInfo(self.target)
-        elif self.plat_name == "win32":
+        if self.plat_name == "win32":
```
davidhewitt, line 82: Regarding the `FIXME`s: how about separating out the crossenv logic into a `get_crossenv_target_info() -> Option[_TargetInfo]`, and calling that here first:

```suggestion
        crossenv_target_info = self.get_crossenv_target_info()
        
        if crossenv_target_info is not None:
            # TODO: Add some kind of check that self.target matches this target info here,
            # raise `ValueError` if mismatches?
            return crossenv_target_info
        
        if self.target:
            return _TargetInfo(self.target)
        elif self.plat_name == "win32":
```

davidhewitt (commented):
Sorry it's taken me so long to review this.

Test looks great to me, thanks for adding! I have a suggestion to resolve the FIXMEs, and then let's merge this.

davidhewitt:
I've added a new commit and rebased to address my last comment; I'll finish this off and merge it 👍

Issue #144, opened 2021-05-04
AttributeError when calling build_rust


akubera:
When running `python setup.py build_rust` in any example, I get the following AttributeError:

```
/workdir/examples/html-py-ever $ python setup.py build_rust
running build_rust
Traceback (most recent call last):
  File "/workdir/examples/html-py-ever/setup.py", line 7, in <module>
    setup(
  File "/workdir/venv/lib/python3.9/site-packages/setuptools/__init__.py", line 165, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python3.9/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib/python3.9/distutils/dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.9/distutils/dist.py", line 985, in run_command
    cmd_obj.run()
  File "/workdir/setuptools_rust/command.py", line 52, in run
    self.run_for_extension(ext)
  File "/workdir/setuptools_rust/build.py", line 138, in run_for_extension
    if self.plat_name.startswith("macosx-") and arch_flags:
AttributeError: 'NoneType' object has no attribute 'startswith'
```

Running `python setup.py build` works as expcted.

It appears that `self.plat_name` is still defined as the [default value None](https://github.com/PyO3/setuptools-rust/blob/main/setuptools_rust/build.py#L61) when `run` is called, so the `self.plat_name.startswith("macosx-")` fails.

Replacing `None` with the empty string fixes the exception, but I think it should still have the correct platform name after `finalize_options`.

It seems that `build` works because it [extracts values](https://github.com/PyO3/setuptools-rust/blob/main/setuptools_rust/setuptools_ext.py#L93) from `bdist_ext`, which [determines `plat_name` by](https://github.com/pypa/setuptools/blob/a4dbe3457d89cf67ee3aa571fdb149e6eb544e88/setuptools/_distutils/command/bdist.py#L92)

```python
if self.plat_name is None:
    if self.skip_build:
        self.plat_name = get_platform()
    else:
        self.plat_name = self.get_finalized_command('build').plat_name
```

Simulating this in `finalize_options()` solves the issue.

https://github.com/PyO3/setuptools-rust/compare/main...akubera:master

I can make a pull request if you'd like, or if you know a better way to fix it, go ahead.

messense:
Feel free to open a pull request.

PR #145, opened 2021-05-05
Add finalization of command option "plat_name"

akubera:
Addresses `AttributeError` raised when running `build_rust`.

Closes issue #144

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -64,6 +64,9 @@ class build_rust(RustCommand):
     def finalize_options(self):
         super().finalize_options()
 
+        if self.plat_name is None:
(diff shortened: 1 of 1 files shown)


messense (approved):
LGTM

gyscos:
Hi!
Could a release be done including this fix? The currently released version seems unable to build most packages.

davidhewitt:
Sorry, I've been busy for a little while. I plan to make a release soon, but there's a few breaking changes on the `main` branch and this will probably be a 1.0 release, so it's going to be a few weeks away.

As a workaround you can just use `setup.py build`, which will then call `build_rust` correctly. This is what most users are doing.

PR #149, opened 2021-06-02
Generate correct tag for universal wheels

luizirber:
Hi!

I'm [moving a Python package](https://github.com/dib-lab/sourmash/pull/1564) from using milksnake to setuptools-rust, since it is more active and it supports `universal2` wheels on macOS. It mostly worked, but I hit a problem related to how milksnake supports [Python's universal wheels](https://github.com/getsentry/milksnake#what-is-supported) (without dependency on the CPython API at all, because `cffi` does all the work). That's how I've been generating wheels and [distributing them at PyPI](https://pypi.org/project/sourmash/4.1.1/#files), but when I generate wheels with `setuptools-rust` it comes out with a `cp38-cp38-linux_x86_64` tag (or `cp38-abi3-linux_x86_64` if using `python setup.py bdist_wheel --py-limited-api=cp38`)

This PR does something similar to what [`milksnake` does](https://github.com/getsentry/milksnake/blob/ef0723e41df23d8f6357570c69c1e69cb31f9e9e/milksnake/setuptools_ext.py#L312L317), but it feels a bit of a hack (especially since there seems to be [some pushback](https://github.com/pypa/setuptools/issues/1976) with universal wheels in general, now that Python 2 is no more).

Is this a change that looks acceptable, or should I look for other approaches to fix my problem?

Thanks!

### DIFF
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -215,6 +215,9 @@ def add_rust_extension(dist):
                         self.bdist_dir,
                         "macosx-{}-universal2".format(macos_target)
                     )
+
+                if self.universal:
+                    return "py3", "none", plat
                 return python, abi, plat
         dist.cmdclass["bdist_wheel"] = bdist_wheel_rust_extension
 


messense:
According to https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels , you should only use `universal = 1` if your project does not have any C extensions and supports Python 2 and 3. I'm not sure about this patch, it could easily get wrong with non-cffi packages?

cc @davidhewitt

messense:
The benefits of universal wheel is that is can run on both Python 2 and 3. Since Python 2 support is no more, if you only needs to support Python 3, using `--py-limited-api=cp3x` where `cp3x` is the minimal supported Python version for example `cp36` should be enough? You only need to distribute a single abi3 wheel for Python 3.

setuptools_rust/setuptools_ext.py
```diff
@@ -215,6 +215,9 @@ def get_tag(self):
                         self.bdist_dir,
                         "macosx-{}-universal2".format(macos_target)
                     )
+
+                if self.universal:
+                    return "py3", "none", plat
```
davidhewitt, line 220: I think the correct `universal` tag is `py2.py3` ?

Also, this should proabably raise a `ValueError` if the binding is not `Binding.NoBinding`, because PyO3 and rust-cpython both depend on the CPython API so anything built with those bindings is definitely not universal.

davidhewitt (commented):
Sorry it has taken me some time to think about this.

Hmm. I understand the point that if your native library doesn't depend on the Python API at all then `py2.py3-none-<platform>` sounds acceptable. Is that actually a wheel extension that `pip` will permit?

It looks like your project actually requires Python 3.7, so you want `py3-none-<platform>`?

davidhewitt:
I'm going to close this as there's been no discussion for over a year, and it's not clear to me that this is a needed feature. If there's still desire, please feel free to comment and move discussion forward.

PR #150, opened 2021-06-20
Patch distutils build command to use `has_ext_modules()`

### LINKED ISSUE #146: Question: `setup.py build` does not put built binary under `build/lib.*`?
I am trying to use poetry with `setuptools-rust` to embed some rust code in to my project.

Poetry "supports" native extensions by specifying a `build.py` scripts that modifies the arguments that's passed into `setuptools.setup`, calls `setup.py build` and then searches native extensions built by the command and adding it to wheel manually.

Currently poetry only searches `build/lib.*` for native extensions after running `setup.py build`. But `setuptools-rust` generates artifacts under `build/lib`.

See: https://github.com/python-poetry/poetry-core/blob/ae3399a5098b8a3ebd0d919ad21a7665e5a8c509/poetry/core/masonry/builders/wheel.py#L146

I have tested some projects with Cython and C extensions (`distutils.core.Extension`) and they produce files under a directory like `build/lib.macosx-10.15-x86_64-3.8` when running `setup.py build`.

I am not very familiar with the intended behaviour of `setup.py build` but the behaviour of this library seem to differ with Cython / C-exts. So I am creating this issue here. If this is the intended behaviour, feel free to close this and I will submit an issue with poetry.

messense:
Fixes #146 

Upstream patch: https://github.com/pypa/distutils/pull/43

### DIFF
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -219,6 +219,25 @@ def add_rust_extension(dist):
         dist.cmdclass["bdist_wheel"] = bdist_wheel_rust_extension
 
 
+def patch_distutils_build():
+    '''Patch distutils to use `has_ext_modules()`
+
+    See https://github.com/pypa/distutils/pull/43
+    '''
+    from distutils.command import build as _build
+
+    class build(_build.build):
+        def finalize_options(self):
+            build_lib_user_specified = self.build_lib is not None
+            super().finalize_options()
+            if not build_lib_user_specified:
+                if self.distribution.has_ext_modules():
+                    self.build_lib = self.build_platlib
+                else:
+                    self.build_lib = self.build_purelib
+
+    _build.build = build
+
 
 def rust_extensions(dist, attr, value):
     assert attr == "rust_extensions"
@@ -229,4 +248,5 @@ def rust_extensions(dist, attr, value):
     )
 
     if dist.rust_extensions:
+        patch_distutils_build()
         add_rust_extension(dist)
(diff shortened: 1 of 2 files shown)


setuptools_rust/setuptools_ext.py
```diff
@@ -219,6 +219,19 @@ def get_tag(self):
...
+def patch_disutils_build():
+    from distutils.command import build as _build
+
+    class build(_build.build):
+        def finalize_options(self):
+            super().finalize_options()
+            if self.distribution.has_ext_modules():
+                self.build_lib = self.build_platlib
+            else:
+                self.build_lib = self.build_purelib
```
davidhewitt, line 231: Doing this will override any user-provided setting for `--build-lib`. Might be possible to avoid that by checking if `--build-lib` set:

```suggestion
            build_lib_user_specified = self.build_lib is not None
            super().finalize_options()
            if not build_lib_user_specified:
                if self.distribution.has_ext_modules():
                    self.build_lib = self.build_platlib
                else:
                    self.build_lib = self.build_purelib
```

setuptools_rust/setuptools_ext.py
```diff
@@ -219,6 +219,19 @@ def get_tag(self):
         dist.cmdclass["bdist_wheel"] = bdist_wheel_rust_extension
 
 
+def patch_disutils_build():
```
davidhewitt, line 222: It would be nice in the source to refer to the upstream `distutils` issue, so that if it's ever merged there's a record of when we might be able to remove this.

davidhewitt (commented):
👍 it looks like this is a necessary trick for now!

PR #165, opened 2021-09-06
docs: fix build-wheels.sh

### LINKED ISSUE #60: Incorrectly pack multiple versions of .so in one wheel
When using build-wheels.sh, in the for-loop of `PYBIN`:

1. When `PYBIN=py35`, `pip wheel` generates `build/lib/xxx.cpython-35m-x86_64-linux-gnu.so` and pack it into `py35.whl`

1. When `PYBIN=py36`, `pip wheel` generates `build/lib/xxx.cpython-36m-x86_64-linux-gnu.so` and pack BOTH `py35.so` and `py36.so` into `py36.whl`

and so on.

Thus, the resulting py38.whl contains all .so files, from py35.so to py38.so.

I don't how setuptools works, maybe the `build/lib` should be cleaned in advance?

davidhewitt:
Updates the `build-wheels.sh` script in the example so that there's only one copy, it's in sync between the `html-py-ever` example and the docs, and also fixes #60


messense (approved):
Cool!

PR #197, opened 2021-12-01
build: don't use abi3 lib suffix when py_limited_api unspecified

davidhewitt:
Something I noticed when using 1.1.0 earlier. Basically with `py_limited_api="auto"` (the default), the abi3 lib name is used by default, which is not correct. The abi3 lib name should only be used with "auto" when `--py-limited-api` is passed to `setup.py bdist_wheel`.

Also added a quick `pytest` job with a unit test for this. Might be useful for later pieces.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,20 @@ jobs:
 
       - run: tox -e mypy
 
+  pytest:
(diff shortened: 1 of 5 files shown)


davidhewitt:
... I'm going to merge immediately and make a 1.1.1 release PR, which this CI can then run in. It's getting rather late 😴

PR #199, opened 2021-12-02
python: support kebab-case

williamstar:


davidhewitt:
Thanks, I'll try to take a look at this soon! My review backlog is currently a few days long, so I'm afraid it might be early next week when I look at this.

williamstar:
👌🏻

davidhewitt:
Thank you for this implementation, which showed me how we can support this. 

When reviewing I decided I prefer a slightly different way for the user to configure kebab-case, by using the `target` dictionary to indicate kebab-case directly. I think that's a slightly more intuitive approach than having the `script` dictionary chain the mapping as proposed here. Opened it as #205

williamstar:
The implementation looks great.

PR #200, opened 2021-12-04
Remove the dependency on tomli to simplify installation

alex:
Depending on tomli produces circular dependencies when attempting to install from sdists only: https://github.com/pyca/cryptography/issues/6671

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.1.2 (UNRELEASED)
+### Changed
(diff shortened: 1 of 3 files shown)


davidhewitt:
Thanks, using `cargo metadata` is a brilliant solution. Reading the cryptography thread, sorry for the pain caused.

I'll follow up this PR tomorrow morning with a quick refactoring so we only need to call `cargo metadata` once (we use it in one other place too) and then put this patch release live.

alex:
Sounds great, thanks!

PR #216, opened 2022-03-20
Add support for custom cargo profiles

### LINKED ISSUE #212: Support for custom profiles
Would be nice of `setuptools-rust` could support custom profiles, which are now stable in Cargo (https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles).

You can't use them in `setuptools-rust` right now via `args=["--profile", "myprofile"]` if `debug=False`, because it conflicts with the added `--release` option. If you set `debug=True` it might actually work, but mostly by accident... and it might have other side-effects that might not be intended by the user, e.g. stripping not working.

messense:
Closes #212 

See also https://github.com/PyO3/maturin/pull/785

### DIFF
--- a/setuptools_rust/extension.py
+++ b/setuptools_rust/extension.py
@@ -170,6 +170,26 @@ class RustExtension:
                 "Can not parse rust compiler version: %s", self.rust_version
             )
 
+    def get_cargo_profile(self) -> Optional[str]:
+        args = self.args or []
+        try:
+            index = args.index("--profile")
+            return args[index + 1]
+        except ValueError:
+            pass
+        except IndexError:
+            raise DistutilsSetupError("Can not parse cargo profile from %s", args)
+
+        # Handle `--profile=<profile>`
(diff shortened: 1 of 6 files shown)


davidhewitt:
Thanks! Think we can test it in CI?

setuptools_rust/extension.py
```diff
@@ -170,6 +170,16 @@ def get_rust_version(self) -> Optional[SimpleSpec]:  # type: ignore[no-any-unimp
                 "Can not parse rust compiler version: %s", self.rust_version
             )
 
+    def get_cargo_profile(self) -> Optional[str]:
+        args = self.args or []
+        try:
+            index = args.index("--profile")
+            return args[index + 1]
```
adamreichold, line 177: The profile can also be passed as `--profile=custom`. Is it worth it trying to handle this?

messense: Thanks for the review, fixed.

adamreichold (approved):

PR #217, opened 2022-03-21
Fix building  macOS arm64 wheel with cibuildwheel

### LINKED ISSUE #206: target selection issue with cibuildwheel when only arm64 or x86_64 is given
👋 I was trying to build discrete x86_64 and arm64 wheels for macos (e.g. `CIBW_ARCHS_MACOS: x86_64 universal2 arm64`), but the the packaged installed from the arm64 wheel fails to load on an M1 mac with the error `(mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64e')`.

Installing the package from the universal2 wheel works fine.

IIUC, this is a compatibility issue with cibuildwheel (we could use `CARGO_BUILD_TARGET` the "manual" way)... To be compatible, the code below should always consider the value of `ARCHFLAGS` to select the target architecture, but it currently only considers it when both x86_64 and arm64 are given (universal2).

https://github.com/PyO3/setuptools-rust/blob/258e198944f555fd85a657bb40b0e5d3f705a306/setuptools_rust/build.py#L90-L109

messense:
Fixes #206

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
   fmt:
     runs-on: "ubuntu-latest"
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


davidhewitt (approved):
Thanks! Would be nice to have a CHANGELOG entry 👍

PR #218, opened 2022-03-22
release: 1.2.0

davidhewitt:
I think we've accumulated enough changes to make worth releasing.

AFAIK it's fine in Python community to drop support for Python 3.6 in a minor release.


messense (approved):
👍

PR #220, opened 2022-04-15
Add `None` default value for `forced_target_triple` and `plat_name`

gabrielmbmb:
The type hinting for the argument `forced_target_triple` of the methods `build_class._detect_rust_target` and `build_class.build_extension` is `Optional[str]`, but it does not have `None` as default value. The same happens whit the class attribute `plat_name`.


davidhewitt:
Thanks. This looks reasonable to me and I believe the CI failure is unrelated. 👍

PR #222, opened 2022-04-23
remove setuptools monkey patch

### LINKED ISSUE #221: TypeError if setuptools_rust are used and python-distutils-extra are installed
I'm sorry if this is the wrong place to report this but I'm not really sure where the problem lies and setuptools_rust just happens to be at the end of traceback ;-).

I've received a bug report from a Gentoo user whose trying to build cryptography (which uses setuptools-rust) on a system where [python-distutils-extra](https://launchpad.net/python-distutils-extra) is installed as well. An attempt to call `build` command results in the following traceback:

```pytb
  File "/tmp/portage/dev-python/cryptography-36.0.2/work/cryptography-36.0.2/setup.py", line 39, in <module>
    setup(
  File "/usr/lib/python3.10/site-packages/setuptools/__init__.py", line 87, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 148, in setup
    return run_commands(dist)
  File "/usr/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 163, in run_commands
    dist.run_commands()
  File "/usr/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 967, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.10/site-packages/setuptools/dist.py", line 1214, in run_command
    super().run_command(command)
  File "/usr/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 985, in run_command
    cmd_obj.ensure_finalized()
  File "/usr/lib/python3.10/site-packages/setuptools/_distutils/cmd.py", line 107, in ensure_finalized
    self.finalize_options()
  File "/usr/lib/python3.10/site-packages/DistUtilsExtra/command/build_extra.py", line 57, in finalize_options
    distutils.command.build.build.finalize_options(self)
  File "/usr/lib/python3.10/site-packages/setuptools_rust/setuptools_ext.py", line 264, in finalize_options
    super().finalize_options()
TypeError: super(type, obj): obj must be an instance or subtype of type
```

Similar problems occur when calling other `setup.py` commands or the PEP517 backend.

davidhewitt:
I believe the upstream PR in https://github.com/pypa/distutils/pull/43 got released in setuptools 58, so we can bump dependency and drop our patching.

@mgorny is there a way in which you can test if this fixes #221? My hunch is that it will.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,12 @@
 
 ## Unreleased
 
+### Packaging
+- Increase minimum `setuptools` version to 1.58. [#222](https://github.com/PyO3/setuptools-rust/pull/222)
+
 ### Fixed
-- Fix building sdist with vendored dependencies on Windows. [#223](https://github.com/PyO3/setuptools-rust/pull/223)
+- Fix crash when `python-distutils-extra` linux package is installed. [#222](https://github.com/PyO3/setuptools-rust/pull/222)
+- Fix sdist built with vendored dependencies on Windows having incorrect cargo config. [#223](https://github.com/PyO3/setuptools-rust/pull/223)
 
 ## 1.2.0 (2022-03-22)
 ### Packaging
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=46.1", "setuptools_scm[toml]>=3.4.3"]
+requires = ["setuptools>=58.0", "setuptools_scm[toml]>=3.4.3"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -240,44 +240,19 @@ def add_rust_extension(dist: Distribution) -> None:
         dist.cmdclass["bdist_wheel"] = bdist_wheel_rust_extension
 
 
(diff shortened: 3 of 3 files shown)


mgorny:
> @mgorny is there a way in which you can test if this fixes #221? My hunch is that it will.

Yes, I can confirm that cryptography builds fine after applying the patch.

examples/rust_with_cffi/noxfile.py
```diff
@@ -7,6 +7,6 @@
 
 @nox.session()
 def test(session: nox.Session):
-    session.install(SETUPTOOLS_RUST, "pytest", "cffi")
+    session.install(SETUPTOOLS_RUST, "pytest", "cffi", "distutils-extra")
```
mgorny, line 10: I don't think python-distutils-extra is on pypi (or at least I couldn't find it). I don't really know nox but if it just wraps pip, I suppose you could use the tarball: `https://launchpad.net/python-distutils-extra/trunk/2.39/+download/python-distutils-extra-2.39.tar.gz`

messense (approved):
👍

examples/rust_with_cffi/noxfile.py
davidhewitt: Ehh, I can't manage to verify the crash with or without that tarball installed locally.

As you've verified the patch works, I'm happy 👍

PR #223, opened 2022-04-23
ci: fix sdist vendor test

davidhewitt:
Trying to debug whatever's started failing in CI...


setuptools_rust/setuptools_ext.py
```diff
@@ -287,3 +279,12 @@ def rust_extensions(
...
         add_rust_extension(dist)
+
+
+_CARGO_VENDOR_CONFIG = b"""
+[source.crates-io]
+replace-with = "vendored-sources"
+
+[source.vendored-sources]
+directory = ".cargo/vendor"
+"""
```
davidhewitt, line 290: @messense the windows CI was failing, I think the number of backslashes changed or something awkward like that. I figured might as well just hard-code it like this. Sound reasonable to you?

messense: Sounds reasonable.

messense (approved):

PR #234, opened 2022-05-31
feat: support cargo --locked flag

### LINKED ISSUE #233: Allow building in locked mode
`cargo` offers [the `--locked` flag](https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options) for reproducibility, which we use in CI to ensure we are building with the exact set of dependencies we have audited.

It would be great to have this option via `setuptools-rust`, to fail building extensions if the relevant lockfiles are out of date.

tommilligan:
Closes #233 

This PR adds support for the `--locked` flag, wherever an external call to `cargo` is issued.

It is important this is supported in all callsites (not just in build), as all cargo calls have the potential to update the lockfile - for instance, `cargo metadata` will do so, which we call before `cargo build` when compiling the extension.

I'm not familiar with the codebase at all, so very happy to be given pointers on if I've missed anything, or the names/default values of the configuration options need to be considered more carefully.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -46,6 +46,7 @@ class build_rust(RustCommand):
         ("debug", "d", "Force debug to true for all Rust extensions "),
         ("release", "r", "Force debug to false for all Rust extensions "),
         ("qbuild", None, "Force enable quiet option for all Rust extensions "),
+        ("locked", None, "Force enable locked option for all Rust extensions "),
         (
             "build-temp",
             "t",
@@ -63,6 +64,7 @@ class build_rust(RustCommand):
         self.debug = None
         self.release = None
         self.qbuild = None
+        self.locked = None
         self.build_temp = None
         self.plat_name = None
         self.target = os.getenv("CARGO_BUILD_TARGET")
@@ -145,9 +147,14 @@ class build_rust(RustCommand):
             target_dir = os.path.join(target_dir, target_triple)
 
         quiet = self.qbuild or ext.quiet
+        locked = self.locked or ext.locked
         debug = self._is_debug_build(ext)
         cargo_args = self._cargo_args(
-            ext=ext, target_triple=target_triple, release=not debug, quiet=quiet
+            ext=ext,
+            target_triple=target_triple,
+            release=not debug,
+            quiet=quiet,
+            locked=locked,
         )
 
         if ext._uses_exec_binding():
@@ -476,6 +483,7 @@ class build_rust(RustCommand):
         target_triple: Optional[str],
         release: bool,
         quiet: bool,
+        locked: bool,
     ) -> List[str]:
         args = []
         if target_triple is not None:
@@ -488,6 +496,9 @@ class build_rust(RustCommand):
 
         if quiet:
             args.append("-q")
+
+        if locked:
+            args.append("--locked")
 
         elif self.verbose:
             # cargo only have -vv
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -46,7 +46,6 @@ class build_rust(RustCommand):
         ("debug", "d", "Force debug to true for all Rust extensions "),
         ("release", "r", "Force debug to false for all Rust extensions "),
         ("qbuild", None, "Force enable quiet option for all Rust extensions "),
-        ("locked", None, "Force enable locked option for all Rust extensions "),
         (
             "build-temp",
             "t",
@@ -64,7 +63,6 @@ class build_rust(RustCommand):
         self.debug = None
         self.release = None
         self.qbuild = None
-        self.locked = None
         self.build_temp = None
         self.plat_name = None
         self.target = os.getenv("CARGO_BUILD_TARGET")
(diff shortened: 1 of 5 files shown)


tommilligan:
Closed as per https://github.com/PyO3/setuptools-rust/issues/233#issuecomment-1142237990

messense (approved):
LGTM

setuptools_rust/build.py
```diff
@@ -46,6 +46,7 @@ class build_rust(RustCommand):
         ("debug", "d", "Force debug to true for all Rust extensions "),
         ("release", "r", "Force debug to false for all Rust extensions "),
         ("qbuild", None, "Force enable quiet option for all Rust extensions "),
+        ("locked", None, "Force enable locked option for all Rust extensions "),
```
davidhewitt, line 49: Given that using setup.py as an entry point is deprecated, I'm kinda 👎 on adding new flags to the `build_rust` command. (I'm not aware of any way to invoke it other than via setup.py, and I wonder if a future setuptools-rust will remove it completely to focus on extending the existing setuptools commands which build frontend expose.)

(I'd quite like to redirect anyone doing `setup.py build_rust` to `maturin`.)

setuptools_rust/extension.py
```diff
@@ -113,6 +114,7 @@ def __init__(
         rustc_flags: Optional[List[str]] = None,
         rust_version: Optional[str] = None,
         quiet: bool = False,
+        locked: bool = False,
```
davidhewitt, line 117: Locked is fine, though as noted should this instead be a more generic `cargo_manifest_args: List[str] = []`?

davidhewitt (commented):
Thanks for this @tommilligan, and sorry for my very slow review. Some things I'd like to see:
- Please add a CHANGELOG entry
- Decide whether to add the new flag to `build_rust`. @messense  I'd be interested in your opinion on this too.
- Consider whether to adopt the more generic `cargo_manifest_args` instead.

If you're willing, it would also be great to add a test for this. Our existing test infrastructure has evolved through time. It might be easiest to use `pytest` to generate a temporary project, try to build with this new `locked` option, and assert the build fails.

messense:
> Decide whether to add the new flag to `build_rust`.

Personally I've never used `build_rust` before, I guess it's for development build? I used to run `build_ext --inplace` for that.

davidhewitt:
> Personally I've never used build_rust before, I guess it's for development build?

Pretty much, it runs the rust build and copies the build library into the right place. It's what we hook setuptools' `build_ext` subcommand to do. Given that `setup.py` as an entrypoint is deprecated, I'd prefer see `setuptools_rust` be simplified to a point where it's intended to used for `pip install .` or `pip install -e .` (or other build frontends), and `maturin` is used for more CLI-heavy tasks.

setuptools_rust/build.py
tommilligan: Removed the option from `build_rust` - it will still use the options if configured on each extension, though.

setuptools_rust/extension.py
tommilligan: Updated to make generic over `locked`, `frozen` and `offline`

tommilligan:
@davidhewitt thanks for your feedback - I've updated to:

- Remove `locked` in favour of a generic `cargo_manifest_args` on extension configuration
- Remove the `build_rust` argument
- Update the changelog

Also rebased on master as there were some conflicts.

I confirmed this works for my use case locally still:

```log
$ python setup.py develop
...
running build_ext
running build_rust
    Updating crates.io index
error: the lock file /home/tom/reinfer/platform/alert/py_test_bridge/Cargo.lock needs to be updated but --locked was passed to prevent this
If you want to try to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
  File "setup.py", line 6, in <module>
    setup(
...
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['cargo', 'metadata', '--manifest-path', 'Cargo.toml', '--format-version', '1', '--locked']' returned non-zero exit status 101.
```

setuptools_rust/setuptools_ext.py
```diff
@@ -48,15 +48,27 @@ def initialize_options(self) -> None:
...
+                # Collate cargo manifest options together.
+                # We can cheat here, as the only valid options are the simple strings
+                # --frozen, --locked, or --offline.
+                #
+                # https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options
+                cargo_manifest_args: Set[str] = set()
                 for ext in self.distribution.rust_extensions:
                     manifest_paths.append(ext.path)
+                    if ext.cargo_manifest_args:
+                        cargo_manifest_args.union(ext.cargo_manifest_args)
```
davidhewitt, line 61: I think this call to `union` doesn't actually update the set?

```python
>>> x = set()
>>> x.union({1, 2})
{1, 2}
>>> x
set()
```

I think you want

```suggestion
                        cargo_manifest_args |= set(ext.cargo_manifest_args)
```

davidhewitt (changes requested):
Thanks again @tommilligan, and sorry for the slow turnaround by me again.

New design looks great to me, just spotted one bug in the implementation I think. If you agree it's a bug, once fixed will be happy to merge this 👍

setuptools_rust/setuptools_ext.py
tommilligan: :facepalm: I blame it on my currently-full-of-covid brain.

Updated to an `update` call, which does a mutable update in place.

davidhewitt (approved):
Looks good to me, thanks for iterating on this together!

PR #240, opened 2022-06-12
Build and test namespace_package wheel for emscripten-wasm32

hoodmane:
I will create separate PRs splitting off the changes to the actual package code.

It seems that the prebuilt emscripten-wasm toolchain doesn't come with relocations. I need to use `-Zbuild-std` to prevent a bunch of errors like
```
relocation R_WASM_MEMORY_ADDR_LEB cannot be used against symbol `__THREW__`; recompile with -fPIC
```
What I do not understand is why I am not needing `-Zbuild-std` to get cryptography to build in
https://github.com/pyodide/pyodide/pull/2683
It's quite mysterious.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -368,7 +368,7 @@ jobs:
       - name: Build namespace wheel
         run: nox -s build-emscripten-namespace-package-wheel
       - name: Test
-        run: nox -s test-emscripten
+        run: nox -s test-emscripten-namespace-package-wheel
 
 
   test-cibuildwheel:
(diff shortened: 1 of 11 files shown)


hoodmane:
If we get Pyodide out of tree builds working, we could simplify this a lot and build and test all four test packages using pytest inside of Pyodide.

davidhewitt:
> It seems that the prebuilt emscripten-wasm toolchain doesn't come with relocations.

Searching for that error some more, I come across https://github.com/rust-lang/rust/issues/60231#issuecomment-656045286. I wonder if that works?

hoodmane:
I'm going to redo this using Pyodide instead of compiling a separate interpreter. I think this will work better.

PR #241, opened 2022-06-12
Fix handling when BLDSHARED has no flags

hoodmane:
If there are no spaces in `bldshared` then `split` returns a list of length 1 and there is an error. OTOH `partition` always returns a tuple of length 3.

Split from #240.


davidhewitt:
Thanks! Could you add a CHANGELOG entry for this please?

PR #243, opened 2022-06-12
Add private environment variable to add extra cargo arguments

hoodmane:
Needed to add `-Zbuild-std` in #240.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -502,6 +502,9 @@ class build_rust(RustCommand):
         if ext.args is not None:
             args.extend(ext.args)
 
+        if "_SETUPTOOLSRUST_EXTRA_CARGO_ARGS" in os.environ:
+            args.append(" ".split(os.environ["_SETUPTOOLSRUST_EXTRA_CARGO_ARGS"]))
+
         return args
 
 


hoodmane:
I'm not sure what the best way to handle this is. It can't be passed in setup.py because it is only needed with certain targets. Maybe you could add a `--cargo-arg` command line argument. Or a `CARGOFLAGS` environment variable.

davidhewitt:
Maybe as an alternative you can create a cargo config with:

```
[unstable]
build-std = true
```

This might avoid needing to add anything to setuptools_rust?

davidhewitt:
Or maybe we could add an API to get the current build target, so that `setup.py` could determine programatically whether to pass the flag?

hoodmane:
> as an alternative you can create a cargo config with ...

Could you explain more? What would the file be named? How would I use it? It's not okay to put in the package's `Cargo.toml` because it depends on the target not on the package.

> Or maybe we could add an API to get the current build target, so that setup.py could determine programatically whether to pass the flag?

The following is almost satisfactory:
```py
extension_args = []
if "CARGO_LOCKED" in os.environ:
	args.append("--locked")

setup(
    name="hello-rust",
    version="1.0",
    rust_extensions=[RustExtension("hello_rust.hello_rust", binding=Binding.PyO3, args=extension_args)],
    packages=["hello_rust"],
    # rust extensions are not zip safe, just like C-extensions.
    zip_safe=False,
)
```
The problem is that it requires the individual packages to be patched to know about the extra argument. If the package isn't aware that we are going to try to build it for emscripten, they won't have code like this. 

I think for composability it would be better to either add an environment variable or a command line argument to add extra cargo flags. IMO an environment variable is the best, like the way `rustc` uses `RUSTFLAGS`.

messense:
It occurred to me that to enable `-Zbuild-std` you might be able to do it with just `CARGO_UNSTABLE_BUILD_STD=true`.

hoodmane:
Grepping the cargo source code doesn't show any instances of environment variables that include `CARGO_UNSTABLE` or or `BUILD_STD` in the names. I still think it would be a good idea to add a `CARGOFLAGS` environment variable that would pass extra arguments to `cargo`.

messense:
> Grepping the cargo source code doesn't show any instances of environment variables that include CARGO_UNSTABLE or or BUILD_STD in the names. 

I think it's handled by a generic [`ConfigKey`](https://github.com/rust-lang/cargo/blob/7a3b56b4860c0e58dab815549a93198a1c335b64/src/cargo/util/config/key.rs#L4-L24)

hoodmane:
Okay I will try it.

hoodmane:
Okay it works. Is this documented anywhere?

messense:
> Okay it works. Is this documented anywhere?

https://doc.rust-lang.org/cargo/reference/config.html#environment-variables

hoodmane:
Oh I guess I was already using that with `CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_LINKER` and `CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER`. I guess it isn't clear from that that `-Zwhatever` turns into `CARGO_UNSTABLE_WHATEVER`. Thanks!

PR #244, opened 2022-06-14
Add CI tests for wasm32-emscripten wheels in Pyodide

hoodmane:
This is another attempt to add emscripten tests. We generate wheels "by hand" and then test them in Pyodide. This approach could benefit from improvements to Pyodide, it's currently a bit ad-hoc. Would still appreciate review.

The tests work for `html-py-ever`, `namespace_package`, and `rust_with_cffi`, but I couldn't get `hello-world` to build so I am not testing it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -412,10 +412,10 @@ jobs:
         with:
           node-version: 18
       - run: |
-          PYODIDE_VERSION=0.21.0-alpha.2
+          PYODIDE_VERSION=0.21.0-alpha.3
 
           cd emscripten
-          npm i pyodide@0.21.0-alpha.2 prettier
+          npm i pyodide@0.21.0-alpha.3 prettier
           cd node_modules/pyodide/
           node ../prettier/bin-prettier.js -w pyodide.asm.js
           EMSCRIPTEN_VERSION=$(node -p "require('./repodata.json').info.platform.split('_').slice(1).join('.')")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - main
-      - emscripten-ci2
   pull_request:
 
 concurrency:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - emscripten-ci2
   pull_request:
 
 concurrency:
@@ -402,3 +403,52 @@ jobs:
         shell: msys2 {0}
         run: |
           PATH="$PATH:/c/Users/runneradmin/.cargo/bin" nox -s test-mingw
+
+  test-emscripten:
+    name: Test Emscripten
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: |
+          PYODIDE_VERSION=0.21.0-alpha.2
+
+          cd emscripten
+          npm i pyodide@0.21.0-alpha.2 prettier
+          cd node_modules/pyodide/
+          node ../prettier/bin-prettier.js -w pyodide.asm.js
+          EMSCRIPTEN_VERSION=$(node -p "require('./repodata.json').info.platform.split('_').slice(1).join('.')")
+          PYTHON_VERSION=3.10.2
+
+          echo "PYODIDE_VERSION=$PYODIDE_VERSION" >> $GITHUB_ENV
+          echo "EMSCRIPTEN_VERSION=$EMSCRIPTEN_VERSION" >> $GITHUB_ENV
+          echo "PYTHON_VERSION=$PYTHON_VERSION" >> $GITHUB_ENV
+          echo "ORIG_PATH=$PATH" >> $GITHUB_ENV
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          components: rust-src
+          target: wasm32-unknown-emscripten
+          override: true
+      - uses: mymindstorm/setup-emsdk@v11
+        with:
+          version: ${{env.EMSCRIPTEN_VERSION}}
+          actions-cache-folder: emsdk-cache
+      - uses: actions/setup-python@v2
+        id: setup-python
+        with:
+          python-version: ${{env.PYTHON_VERSION}}
+      - run: pip install nox
+      - uses: actions/cache@v3
+        with:
+          path: |
+            tests/pyodide
+          key: ${{ hashFiles('tests/*.js') }} - ${{ hashFiles('noxfile.py') }} - ${{ steps.setup-python.outputs.python-path }}
+      - uses: Swatinem/rust-cache@v1
+      - name: Test
+        run: |
+          export PATH=$ORIG_PATH:$PATH
+          nox -s test-examples-emscripten
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import tarfile
 from glob import glob
 from pathlib import Path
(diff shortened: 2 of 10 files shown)


hoodmane:
I'm keeping this as a draft because IMO it needs some improvements before it can merge but I would still appreciate feedback if you get around to it. @messense @davidhewitt

noxfile.py
```diff
@@ -35,3 +36,69 @@ def mypy(session: nox.Session):
...
+        "packages.json",
+        "pyodide.asm.js",
+        "pyodide.asm.data",
+        "pyodide.asm.wasm",
+        "pyodide_py.tar",
+    ]
+    with session.chdir(pyodide_dir):
+        for file in pyodide_files:
+            session.run("wget", PYODIDE_DEV + file, external=True)
+        session.run("npm", "i", "node-fetch")
```
messense, line 56:
```suggestion
        session.run("npm", "i", "node-fetch", external=True)
```

noxfile.py
```diff
@@ -35,3 +36,69 @@ def mypy(session: nox.Session):
...
+            RUSTUP_TOOLCHAIN="nightly",
+            _SETUPTOOLSRUST_BUILD_STD="1",
+            PYTHONPATH=str(emscripten_dir),
+            _PYTHON_SYSCONFIGDATA_NAME="_sysconfigdata__emscripten_wasm32-emscripten",
+            _PYTHON_HOST_PLATFORM="emscripten_3_1_13_wasm32",
+            CARGO_BUILD_TARGET="wasm32-unknown-emscripten",
+            CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_LINKER=str(
+                emscripten_dir / "emcc_wrapper.py"
+            ),
+            PYO3_CONFIG_FILE=str(emscripten_dir / "pyo3_config.ini"),
```
davidhewitt, line 90: From this should I infer that `pyo3_build_config` doesn't configure for pyodide correctly? Do you think we would be able to upstream a fix?

noxfile.py
```diff
@@ -35,3 +36,69 @@ def mypy(session: nox.Session):
...
+            RUSTFLAGS=" ".join(
```
davidhewitt, line 91: It might be worth having a .cargo/config.toml with a `[target.wasm32-unknown-emscripten]` section for these? Or alternatively `setuptools-rust` could set them automatically internally?

setuptools_rust/build.py
```diff
@@ -505,6 +505,9 @@ def _cargo_args(
         if ext.args is not None:
             args.extend(ext.args)
 
+        if "_SETUPTOOLSRUST_BUILD_STD" in os.environ:
```
davidhewitt, line 508: I guess we learned we don't need this and can just have `CARGO_UNSTABLE_BUILD_STD=true`.

emscripten/pyconfig.h
```diff
@@ -0,0 +1,1740 @@
+/* pyconfig.h.  Generated from pyconfig.h.in by configure.  */
```
davidhewitt, line 1: I'm curious what this file is used for?

davidhewitt (commented):
Thanks for working on this and sorry for the long delay in review. Been busy either side of falling ill.

Looks solid to me, really cool to see this working. There appears to be a `rust_with_cffi` failure when trying to build `cffi`, maybe it makes sense to skip that one for now?

hoodmane:
This can be simplified quite a lot now similar to this PR on maturin:
https://github.com/PyO3/maturin/pull/983/files

emscripten/pyconfig.h
hoodmane: It's only needed for `rust_with_cffi` which doesn't work because I didn't get the headers quite right anyways. I'm going to work on getting Pyodide out-of-tree builds working which should give us a standardized way to get the right headers.

noxfile.py
hoodmane: > From this should I infer that pyo3_build_config doesn't configure for pyodide correctly?

I am not sure. I think it may just be that I don't know how to use the tool correctly. How would I check this?

noxfile.py
hoodmane: Well we are working on upstreaming this stuff:
https://github.com/rust-lang/rust/pulls?q=author%3Ahoodmane
I don't know what the best thing to do before the upstream fixes go in. Unfortunately it looks like the merge window for Rust 1.63.0 is closing today. The fixes will probably go into into 1.64.0.

davidhewitt: Ah, very nice. I think it's ok for us to keep this as-is for now then; if things will eventually "just work" on a new enough rustc I think that's good enough for now.

noxfile.py
davidhewitt: So usually in cross-compile situations like for emscripten the `pyo3_build_config` build scripts will try to look for a sysconfigdata file and configure from that. Alternatively, can set the `abi3-py310` feature and then it will do a "standard" cross compile, which might just work.

I view the `PYO3_CONFIG_FILE` env var as an escape hatch when nothing built-in works, so if either of the above work (or can be made to work) that's probably preferable.

.github/workflows/ci.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - emscripten-ci2
```
davidhewitt, line 6: Reminder to remove this when ready to merge.

noxfile.py
```diff
@@ -35,3 +35,51 @@ def mypy(session: nox.Session):
...
+    for example in test_crates:
+        env = os.environ.copy()
+        env.update(
+            RUSTUP_TOOLCHAIN="nightly",
+            _SETUPTOOLSRUST_BUILD_STD="1",
+            PYTHONPATH=str(emscripten_dir),
+            _PYTHON_SYSCONFIGDATA_NAME="_sysconfigdata__emscripten_wasm32-emscripten",
+            _PYTHON_HOST_PLATFORM="emscripten_3_1_14_wasm32",
+            CARGO_BUILD_TARGET="wasm32-unknown-emscripten",
+            CARGO_UNSTABLE_BUILD_STD="true",
```
davidhewitt, line 68: Looking at the CI failure and [docs for `build-std`](https://doc.rust-lang.org/cargo/reference/unstable.html#build-std) I think this actually might want to be `CARGO_UNSTABLE_BUILD_STD="core,alloc,std,proc_macro,test"`. Or presumably we can drop this soon-ish when your upstream Rust PRs merge.

davidhewitt (commented):
Looks really tidy! Think I've spotted why the new test failure...

davidhewitt:
Rebased this and removed some of the tweaks which I think are now covered by upstream rustc; let's see if we can get this merged.

davidhewitt:
Hmm, emscripten failing with missing symbols. @hoodmane do you know if there's a flag for allowing the undefined symbols through (presumably they are provided by Pyodide at runtime), similar to `-undefined dynamic_lookup` which we have to use on macOS?

hoodmane:
I suspect it didn't see `-sSIDE_MODULE=2` for some reason.

hoodmane:
Yeah I think `-sWASM_BIGINT` had better stay too.

.github/workflows/ci.yml
```diff
@@ -402,3 +402,52 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: |
+          PYODIDE_VERSION=0.21.0-alpha.2
+
+          cd emscripten
+          npm i pyodide@0.21.0-alpha.2 prettier
```
messense, line 418:
```suggestion
          PYODIDE_VERSION=0.21.0-alpha.3

          cd emscripten
          npm i pyodide@0.21.0-alpha.3 prettier
```

davidhewitt:
Finally working!

davidhewitt (approved):
Thanks for upgrading the version @messense!

messense (approved):
👍

hoodmane:
Thanks @messense and @davidhewitt!

davidhewitt:
Cheers, appreciate you kicking this off for us, I've got a much better idea how we do (pyodide) support now going forward 👍

PR #246, opened 2022-06-14
Switch from `distutils.command.build` to `setuptools.command.build`

messense:
https://github.com/pypa/setuptools/pull/3256


davidhewitt (commented):
👍, though needs a CHANGELOG entry mentioning the setuptools version bump please!

messense:
Looks like crossenv doesn't support setuptools>=61.

https://github.com/conda-forge/crossenv-feedstock/blob/main/recipe/patches/gh94.patch

PR #248, opened 2022-06-17
Add support for packaging binaries in scripts data directory

### LINKED ISSUE #247: Support packaging binaries in `.data` directory
Currently `Exec` binding with `script=True` are installed as console scripts. It'd be nice to support packaging binaries directly in `.data/scripts` directory as specified by https://peps.python.org/pep-0491/#the-data-directory

messense:
Closes #247

### DIFF
--- a/setuptools_rust/extension.py
+++ b/setuptools_rust/extension.py
@@ -1,6 +1,7 @@
 import json
 import os
 import re
+import warnings
 import subprocess
 from distutils.errors import DistutilsSetupError
 from enum import IntEnum, auto
@@ -160,6 +161,12 @@ class RustExtension:
 
         self._cargo_metadata: Optional[_CargoMetadata] = None
 
+        if binding == Binding.Exec and script:
+            warnings.warn(
+                "'Binding.Exec' with 'script=True' is deprecated, use 'RustBin' instead.",
+                DeprecationWarning,
+            )
+
     def get_lib_name(self) -> str:
         """Parse Cargo.toml to get the name of the shared library."""
         metadata = self._metadata()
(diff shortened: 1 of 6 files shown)


setuptools_rust/extension.py
```diff
@@ -195,9 +195,10 @@ def entry_points(self) -> List[str]:
         entry_points = []
         if self.script and self.binding == Binding.Exec:
             for executable, mod in self.target.items():
-                base_mod, name = mod.rsplit(".")
-                script = "%s=%s.%s:run" % (name, base_mod, _script_name(executable))
-                entry_points.append(script)
+                if "." in mod:
```
messense, line 198: I don't like this, I'm thinking about change `script=False` to this the behavior of this PR. 

IMO the current `Binding.Exec` with `script=False` build mode doesn't not much make sense since it only build the binary and put in somewhere in the package.

messense:
Hopefully this will resolve https://github.com/conda-forge/maturin-feedstock/pull/60#issuecomment-1155728686

setuptools_rust/extension.py
davidhewitt: Trying to get my head around this comment. I agree that `script=False` is kind of weird / useless at the moment. Is there genuinely a use case where the current behaviour of placing the binary into `site-packages` is desirable?

On the other hand, if we change `script=False` to match this PR - I understand that this would mean that the binary would by default be copied into the `bin` directory (and added to `PATH`, essentially). Once we have that, is there ever a reason to want to use `script=True`?

messense: IMO most users reach to `Exec` binding and `script=True` to put the binary in `PATH`, otherwise it's not very useful.

In maturin we also build `bin` bindings by putting it in script data directory directly.

davidhewitt: So if we change `script=False` to work like maturin is there a reason to use `script=True` ever? Maybe we can simplify and deprecate the argument?

messense: > So if we change script=False to work like maturin is there a reason to use script=True ever?

I don't think there is any.

> Maybe we can simplify and deprecate the argument?

I think so.

According to the docs, `target` accepts `Union[str, Dict[str, str]]` right now

> target `(Union[str, Dict[str, str]])` – The full Python dotted name of the extension, including any packages, i.e not a filename or pathname. It is possible to specify multiple binaries, if extension uses Binding.Exec binding mode. In that case first argument has to be dictionary. Keys of the dictionary correspond to the rust binary names and values are the full dotted name to place the executable inside the python package. To install executables with kebab-case names, the final part of the dotted name can be in kebab-case. For example, hello_world.hello-world will install an executable named hello-world.

I'd also like to change it to accept `List[str]` and deprecate `Dict[str, str]` for multiple binary names.

davidhewitt: I wonder what if we went further and made a `RustBin` class to use like `RustExtension`? Then no need to accept a `List` at all I think, users can write the list at the `setup()` level.

messense: A `RustBin` class is nice, I'll give it a try.

davidhewitt: (I'm thinking that way maybe we can leave the existing behavior and emit warnings for deprecation, and then the `RustBin` class can have the new behavior which we want to have going forward. Gives users time to tell us they need the old way before we break / remove.)

messense: Done.

davidhewitt (approved):
LGTM, thanks! Looks like needs rebase for CHANGELOG conflict, feel free to merge after 👍

davidhewitt:
Hmmm, `test_cross` seems to be hanging?

messense:
> Hmmm, `test_cross` seems to be hanging?

cross released [0.2.2](https://github.com/cross-rs/cross/releases/tag/v0.2.2) recently, something might be broken, I'm going to pin it to 0.2.1 for now and investigate later.

messense:
> Hmmm, `test_cross` seems to be hanging?

Opened https://github.com/cross-rs/cross/issues/864

PR #251, opened 2022-06-26
Fix `test_cross` with cross 0.2.2

messense:
Workaround https://github.com/cross-rs/cross/issues/864


messense:
The second commit works around https://github.com/cross-rs/cross/issues/863.

PR #254, opened 2022-06-27
capture errors from `cargo metadata`

### LINKED ISSUE #253: Call to `cargo metadata` doesn't check for errors
In https://github.com/PyO3/setuptools-rust/blob/v1.3.0/setuptools_rust/extension.py#L225, the call to `cargo metadata` does not check for errors of the subprocess call, in contrast to how cargo is [invoked](https://github.com/PyO3/setuptools-rust/blob/v1.3.0/setuptools_rust/build.py#L212-L217) for `cargo build`, for example.

This makes debugging harder, because the result (e.g. in https://github.com/conda-forge/cryptography-feedstock/pull/100) looks like:
```
    File "[...]/lib/python3.7/site-packages/setuptools_rust/extension.py", line 225, in _metadata
      self._cargo_metadata = json.loads(subprocess.check_output(metadata_command))
    File "[...]/lib/python3.7/subprocess.py", line 411, in check_output
      **kwargs).stdout
    File "[...]/lib/python3.7/subprocess.py", line 512, in run
      output=stdout, stderr=stderr)
  subprocess.CalledProcessError: Command '['cargo', 'metadata', '--manifest-path', 'src/rust/Cargo.toml', '--format-version', '1']' died with <Signals.SIGKILL: 9>.
  error: subprocess-exited-with-error
```

With the lack of trace from the subprocess, it's not even possible to add a `export CARGO_LOG=trace`, because it all just gets eaten (AFAICT).

h-vetinari:
Should fix #253.

The setup is copied largely from what's already in [`build.py`](https://github.com/PyO3/setuptools-rust/blob/v1.3.0/setuptools_rust/build.py#L212-L217), but using `DistutilsSetupError` instead of `CompileError` (in line with the rest of the error reporting in `extension.py`). Also, I don't think duplicating the `_prepare_build_environment` is necessary just for `cargo metadata`...?

### DIFF
--- a/setuptools_rust/extension.py
+++ b/setuptools_rust/extension.py
@@ -240,7 +240,25 @@ class RustExtension:
             ]
             if self.cargo_manifest_args:
                 metadata_command.extend(self.cargo_manifest_args)
-            self._cargo_metadata = json.loads(subprocess.check_output(metadata_command))
+
+            try:
+                payload = subprocess.check_output(
+                    metadata_command, encoding="latin-1", stderr=subprocess.PIPE
+                )
+            except subprocess.CalledProcessError as e:
+                raise DistutilsSetupError(
+                    f"""
+                    cargo metadata failed with code: {e.returncode}
+
+                    Output captured from stderr:
+                    {e.stderr}
+
+                    Output captured from stdout:
+                    {e.stdout}
+                    """
+                )
+
+            self._cargo_metadata = json.loads(payload)
         return self._cargo_metadata
 
     def _uses_exec_binding(self) -> bool:
--- a/setuptools_rust/extension.py
+++ b/setuptools_rust/extension.py
@@ -257,8 +257,15 @@ class RustExtension:
                     {e.stdout}
                     """
                 )
-
-            self._cargo_metadata = json.loads(payload)
+            try:
+                self._cargo_metadata = json.loads(payload)
+            except json.decoder.JSONDecodeError as e:
+                raise DistutilsSetupError(
+                    f"""
+                    Error parsing output of cargo metadata as json; received:
+                    {payload}
+                    """
+                ) from e
         return self._cargo_metadata
 
     def _uses_exec_binding(self) -> bool:
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -210,9 +210,21 @@ class build_rust(RustCommand):
 
         # Execute cargo
         try:
-            output = subprocess.check_output(command, env=env, encoding="latin-1")
+            output = subprocess.check_output(
+                command, env=env, encoding="latin-1", stderr=subprocess.PIPE
+            )
         except subprocess.CalledProcessError as e:
-            raise CompileError(f"cargo failed with code: {e.returncode}\n{e.output}")
+            raise CompileError(
+                f"""
+                cargo failed with code: {e.returncode}
+
+                Output captured from stderr:
+                {e.stderr}
+
+                Output captured from stdout:
+                {e.stdout}
+                """
+            )
 
         except OSError:
             raise DistutilsExecError(


setuptools_rust/extension.py
```diff
@@ -240,7 +240,17 @@ def _metadata(self) -> "_CargoMetadata":
             ]
             if self.cargo_manifest_args:
                 metadata_command.extend(self.cargo_manifest_args)
-            self._cargo_metadata = json.loads(subprocess.check_output(metadata_command))
+
+            env = os.environ.copy()
+            try:
+                payload = subprocess.check_output(
+                    metadata_command, env=env, encoding="latin-1"
```
davidhewitt, line 247: The default is to inherit the current environment? I don't think `env=env` is necessary?

setuptools_rust/extension.py
```diff
@@ -240,7 +240,17 @@ def _metadata(self) -> "_CargoMetadata":
...
+                )
+            except subprocess.CalledProcessError as e:
+                msg = f"cargo failed with code: {e.returncode}\n{e.output}"
```
davidhewitt, line 250: Should this also include stderr? Also it would be nice to say `cargo metadata` instead of just `cargo`.

davidhewitt (commented):
Thanks for the PR! Bit of feedback...

setuptools_rust/extension.py
h-vetinari: Fair point, I just wanted to make sure `env` definitely is present, but will remove.

setuptools_rust/extension.py
h-vetinari: As I wrote, I copied the setup from `build.py`. Happy to add `e.stderr` here (and also in `build.py` if desired)

h-vetinari:
Thanks for the quick review!

setuptools_rust/extension.py
h-vetinari: Made these adaptations now.

setuptools_rust/extension.py
```diff
@@ -240,7 +240,23 @@ def _metadata(self) -> "_CargoMetadata":
             ]
             if self.cargo_manifest_args:
                 metadata_command.extend(self.cargo_manifest_args)
-            self._cargo_metadata = json.loads(subprocess.check_output(metadata_command))
+
+            try:
+                payload = subprocess.check_output(metadata_command, encoding="latin-1")
```
davidhewitt, line 245: Ah, I'm really sorry to have missed on the first read, I think we might need to opt-in to capturing stderr:

```suggestion
                payload = subprocess.check_output(metadata_command, encoding="latin-1", stderr=subprocess.PIPE)
```

davidhewitt: (could also use `stderr=subprocess.STDOUT` to merge the two together so that the error could just use `e.stdout`, I think both merged and separate have advantages and disadvantages so happy to go with whichever you prefer)

h-vetinari: Good point, thanks!

I preferred the separate versions (no reliance that merge works correctly/legibly), but happy to change if desired.

h-vetinari: Ah, never mind, the merging is apparently the recommended (or at least documented way): https://docs.python.org/3/library/subprocess.html#subprocess.check_output

h-vetinari: As an added benefit, this makes the exception message (& diff) much smaller.  :)

h-vetinari:
I struggle to understand how any of the recent commits could make the json parsing fail (and the CI passed for) https://github.com/PyO3/setuptools-rust/pull/254/commits/b6752c8219b0325a7500b42d7bbf3815f4dc4356...

setuptools_rust/extension.py
h-vetinari: Well, the merge was what's causing the json errors... It adds the logs into the output, making the json unparseable, e.g.:
```
Error parsing output of cargo metadata as json; received:
                          b'    Updating crates.io index\n Downloading crates ...\n  Downloaded unindent v0.1.8\n  Downloaded target-lexicon v0.12.3\n  Downloaded pyo3-macros v0.16.5\n  Downloaded cfg-if v1.0.0\n  Downloaded scopeguard v1.1.0\n  Downloaded unicode-xid v0.2.2\n  Downloaded proc-macro2 v1.0.36\n  Downloaded indoc v1.0.4\n  Downloaded instant v0.1.12\n  Downloaded once_cell v1.10.0\n  Downloaded lock_api v0.4.6\n  Downloaded parking_lot v0.11.2\n  Downloaded bitflags v1.3.2\n  Downloaded pyo3-ffi v0.16.5\n  Downloaded quote v1.0.16\n  Downloaded syn v1.0.89\n  Downloaded pyo3-macros-backend v0.16.5\n  Downloaded smallvec v1.8.0\n  Downloaded redox_syscall v0.2.11\n  Downloaded parking_lot_core v0.8.5\n  Downloaded pyo3 v0.16.5\n  Downloaded winapi v0.3.9\n  Downloaded libc v0.2.121\n  Downloaded pyo3-build-config v0.16.5\n  Downloaded winapi-x86_64-pc-windows-gnu v0.4.0\n  Downloaded winapi-i686-pc-windows-gnu v0.4.0\n{"packages":[{"name":"bitflags",
```

davidhewitt: Ah, I see!

Now that I realise that these updating / downloading messages come through stderr - it's probably quite nice to get them as output by default (i.e. not capture stderr)? 

Could maybe check the `--quiet` flag - if set, capture and include `stderr` in the exception message, otherwise just don't capture stderr? What do you think of that?

I _think_ we're pretty much there now - thanks for the patience / iteration on this!

h-vetinari: I went back to trying the non-merged version for now. :)

h-vetinari: This is looking better now. :)

Issue #235, opened 2022-05-31
rustc_flags applied only to top-level crate


cassiersg:
When `rustc_flags` are given to `RustExtension`, these flags are used only to build to top-level crate, and not the dependencies (at least for cdylib builds). This is due to using `cargo rustc` as build command, instead of using `cargo build` while setting the `RUSTFLAGS` environment variable (see https://doc.rust-lang.org/cargo/commands/cargo-rustc.html).

Is this behavior a bug ? If not, would you welcome a PR for an additional flag that would be applied to all crates built ?

As an example, my use-case is the flag `-C target-feature=+avx2`, which has to be applied to the dependencies.

davidhewitt:
Hi, sorry for the slow reply. I agree this is a bug, a PR would be very welcome!

davidhewitt:
I had a further thought about this and decided that it's possible there are valid use cases for `rustc_flags`, so I decided against changing them.

In #260 I documented this behavior and suggest using `RUSTFLAGS` as above. If an environment variable is inconvenient you can always have a `.cargo/config` file to set flags.

PR #260, opened 2022-07-02
clarify semantics of `rustc_flags`, deprecate `native`

davidhewitt:
I had a think about #235 and decided that it's not actually a bug, it's intended behaviour. The documentation now says you should use `RUSTFLAGS` environment variable if you want to modify the whole build.

Despite this, I did tidy up some of our internal flags like `-C linker=...` and the musl flags. Also added a `print` statement so that users can see what the final `RUSTFLAGS` are set to (if there are any and not `quiet`).

Finally, deprecated the `native` option in favour of letting users just configure this themselves with flags. I generally think it's not very useful as using `native` will produce extensions which may be less portable. So I don't think library authors will typically use it. 

Once this is merged (if it passes review), let's put out 1.4.0 release.

### DIFF
--- a/setuptools_rust/extension.py
+++ b/setuptools_rust/extension.py
@@ -5,7 +5,7 @@ import subprocess
 import warnings
 from distutils.errors import DistutilsSetupError
 from enum import IntEnum, auto
-from typing import Any, Dict, List, NewType, Optional, Union
+from typing import Any, Dict, List, NewType, Optional, Sequence, Union
 
 from semantic_version import SimpleSpec
 from typing_extensions import Literal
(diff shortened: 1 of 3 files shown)


setuptools_rust/extension.py
```diff
@@ -280,46 +284,45 @@ class RustBin(RustExtension):
...
 
     def __init__(
         self,
-        target: str,
+        target: Union[str, Dict[str, str]],
         path: str = "Cargo.toml",
-        args: Optional[List[str]] = None,
-        cargo_manifest_args: Optional[List[str]] = None,
-        features: Optional[List[str]] = None,
-        rustc_flags: Optional[List[str]] = None,
```
davidhewitt, line 303: Deliberately removed `rustc_flags` from `RustBin` because it uses `cargo build` so never passes these flags.

messense (approved):

PR #256, opened 2022-06-28
respect `quiet` flag in `cargo metadata`

davidhewitt:
Follow up to #254 to make the formatting reusable & tested, and respect `quiet` flag for stderr.

PR #265, opened 2022-07-06
correctly apply quiet-flag to listen to stderr or not

h-vetinari:
Opening this for discussion (which is harder to do in closed issues & PRs).

Even with the brandnew, 1.4.0 I still do not get the logmessages (e.g. from `export CARGO_LOG=trace`) from the crashing `cargo metadata` call, c.f. #253. I think the condition in #256 ended up being the wrong way around - this tries to fix that.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -207,7 +207,7 @@ class build_rust(RustCommand):
 
         # Execute cargo
         try:
-            stderr = subprocess.PIPE if quiet else None
+            stderr = None if quiet else subprocess.PIPE
             output = subprocess.check_output(command, env=env, stderr=stderr, text=True)
         except subprocess.CalledProcessError as e:
             raise CompileError(format_called_process_error(e))
--- a/setuptools_rust/extension.py
+++ b/setuptools_rust/extension.py
@@ -232,7 +232,7 @@ class RustExtension:
                 f.write(_SCRIPT_TEMPLATE.format(executable=repr(executable)))
 
     def _metadata(self, *, quiet: bool) -> "_CargoMetadata":
-        """Returns cargo metedata for this extension package.
+        """Returns cargo metadata for this extension package.
 
         Cached - will only execute cargo on first invocation.
         """
@@ -249,7 +249,7 @@ class RustExtension:
                 metadata_command.extend(self.cargo_manifest_args)
 
             try:
-                stderr = subprocess.PIPE if quiet else None
+                stderr = None if quiet else subprocess.PIPE
                 payload = subprocess.check_output(
                     metadata_command, stderr=stderr, encoding="latin-1"
                 )


setuptools_rust/build.py
```diff
@@ -207,7 +207,7 @@ def build_extension(
 
         # Execute cargo
         try:
-            stderr = subprocess.PIPE if quiet else None
+            stderr = None if quiet else subprocess.PIPE
```
davidhewitt, line 210: I'm pretty sure the original condition was what I intended. My understanding was:

- if `quiet` we want to capture all output from cargo, so we use `subprocess.PIPE`
- otherwise, pass `None` so that `cargo` will inherit the stderr handle from this process

davidhewitt (commented):
Hmm, it's still confusing why `CARGO_LOG=trace` is not working for you.

I think my original condition worked as intended - see other comment.

To help debug your system - we support overriding `cargo` with the `CARGO` environment variable. You could try writing some scripts that do stuff like print their environment to stderr and then see what they do when invoked here?

setuptools_rust/build.py
adamreichold: I think there might be some confusion w.r.t. the `if not quiet:` clause further below. By using `stderr=subprocess.PIPE` with `check_output`, the stderr will essentially be discarded as `check_output` explicitly returns only `stdout`. Maybe `stderr=subprocess.STDOUT` to merge stderr into stdout and print both in line 221 would be more helpful?

h-vetinari: > * if `quiet` we want to capture all output from cargo, so we use `subprocess.PIPE`

Do you mean capture-and-discard the output from `cargo`? Because capture-and-show would be the opposite of quiet.

I admit that I don't understand the setup of `check_output` well, but the way I read the [docs](https://docs.python.org/3/library/subprocess.html#subprocess.check_output) is that the `stderr=PIPE` is actually necessary to capture the output in the first place (whether it is then discarded or shown), in particular: 

> This is equivalent to:
>
> ```run(..., check=True, stdout=PIPE).stdout```

and `subprocess.run` not capturing stderr by default.

h-vetinari: When I tried subprocess.STDOUT in #254, it [corrupted](https://github.com/PyO3/setuptools-rust/pull/254#discussion_r908114339) the JSON result of `cargo metadata`.

PS. Sorry, your first comment hadn't shown up in my open tab before I responded.

h-vetinari: @davidhewitt @adamreichold 
Any ideas on how to move forward on this? I think we can agree that how stdin/stout are handled are somewhat confusing. Perhaps it might make sense to not use the `check_output` wrapper, but rather `run` directly, so we have more explicit control (or at least clearer documentation)?

davidhewitt: @h-vetinari what do you think of #268?

h-vetinari: Fine by me, thanks 👍 

Would it make sense to add a test that switches on `CARGO_LOG=trace` and expects a minimum number of lines of output?

davidhewitt: Sure thing, added in https://github.com/PyO3/setuptools-rust/pull/268/commits/6eb0ab89fa561eb21097f4cd71e038de2e1635cb

davidhewitt:
Closing as superseded by #268

PR #267, opened 2022-07-10
add mingw CI

davidhewitt:


davidhewitt:
Hmm I think this isn't picking up the right Python

.github/workflows/ci.yml
```diff
@@ -350,3 +350,41 @@ jobs:
...
+        with:
+          profile: minimal
+          toolchain: stable
+          target: ${{ matrix.rust-target }}
+
+      - name: Install test dependencies
+        run: pip install --upgrade nox
+
+      - name: Test examples
+        shell: bash
```
messense, line 387: I think you need

```suggestion
        shell: msys2 {0}
```

See https://github.com/msys2/setup-msys2#usage

davidhewitt:
Hmm seems like `nox` is misbehaving

.github/workflows/ci.yml
davidhewitt: Thanks!

.github/workflows/ci.yml
```diff
@@ -350,3 +350,46 @@ jobs:
...
+            mingw-w64-${{ matrix.arch }}-python
+            mingw-w64-${{ matrix.arch }}-python-pip
+            mingw-w64-${{ matrix.arch }}-openssl
+            mingw-w64-${{ matrix.arch }}-toolchain
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          target: ${{ matrix.rust-target }}
```
messense, line 382: I think we should install Rust in msys2 shell or build it with `--target x86_64-pc-windows-gnu`.

Looks like it's using `x86_64-pc-windows-msvc` because it's the default toolchain.

messense, line 382:
```suggestion
          toolchain: stable-${{ matrix.rust-target }}
          default: true
```

.github/workflows/ci.yml
```diff
@@ -350,3 +350,46 @@ jobs:
...
+            git
+            mingw-w64-${{ matrix.arch }}-python
+            mingw-w64-${{ matrix.arch }}-python-pip
+            mingw-w64-${{ matrix.arch }}-openssl
+            mingw-w64-${{ matrix.arch }}-toolchain
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable-${{ matrix.rust-target }}
```
messense, line 381:
```suggestion
          toolchain: stable-${{ matrix.rust_target }}
```

messense:
> error: Rust build failed; unable to find any *html_py_ever.so in D:\a\setuptools-rust\setuptools-rust\examples\html-py-ever\target/release

Reproduced the issue.

davidhewitt:
Nice work @messense!

Hmm I tried to use `tmate` action to get a debugging shell but I can't use it for some reason. Worked earlier, will try again later. 

I think that the likely code is here: https://github.com/PyO3/setuptools-rust/blob/4c88c6aa5460b1e7146e95316ca6a85f0e61252c/setuptools_rust/build.py#L261

I think we probably need to check `sys.platform` on msys2 and then fixup that condition. Hence why I wanted a debug shell :)

davidhewitt:
Sounds like from https://github.com/pypa/virtualenv/issues/1338#issuecomment-485425589 that `sys.platform` might also be `msys` or maybe `msys2`.

To be honest I think the artifact suffix is really something controlled by rustc, so I'll refactor this condition to be based on the rust target.

davidhewitt:
@messense so I've got the mingw builds working by reading artifacts from the cargo output messages. This is probably more robust than the previous implementation when cross-compiling too.

However hit a couple of breaking problems. Wonder if you've seen these in maturin & might know how to work around:
- `cross` uses different paths in the metadata so the copy fails (I think)
- `cargo-zigbuild` doesn't support `cargo-zigbuild metadata`. Unsure if it should?

(`pytest` job expected to fall, it's from the debug prints I've added.)

messense:
> cargo-zigbuild doesn't support cargo-zigbuild metadata. Unsure if it should?

I can add `cargo-zigbuild metadata`, it'd just delegate to `cargo metadata` since metadata command does not require special setup when cross compiling.

messense:
> Copying rust artifact from /target/aarch64-unknown-linux-gnu/release/libnamespace_package_rust.so to build/lib.linux-x86_64-cpython-38/namespace_package/rust.so

cross use docker containers, so it's the path in container not the host.

davidhewitt:
Fair enough! I seem to have reached an approach which works now by adjusting the target directory if using cross.

messense (approved):
👍

PR #271, opened 2022-07-27
Install cross from prebuilt binary

messense:
Reduce CI time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,7 +260,9 @@ jobs:
           toolchain: stable
           override: true
       - name: Install cross
-        run: cargo install cross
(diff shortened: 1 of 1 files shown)


messense:
https://github.com/PyO3/setuptools-rust/blob/8fe44b93b201926dc71ef8e87ef8e963b9d996e1/setup.cfg#L3

Hmm, looks like this breaks the build in newer setuptools version.

messense:
I wonder if it's desirable to just replace `setuptools-scm` with [bumpversion](https://github.com/c4urself/bump2version). It much simpler.

davidhewitt:
I would be fine with that, would definitely be less brittle 👍

Issue #259, opened 2022-07-02
2.0 wishlist


davidhewitt:
Just some tidy-ups I'd like to do whenever we get around to a next breaking release:

- Change `Optional[Sequence[str]]` arguments to `RustExtension` to just be `Sequence[str] = ()`

jameshilliard:
> * Remove `build_rust` subcommand (the `setup.py` entry point is deprecated, so no need to support it IMO)

I would recommend holding off on this as [it's not currently possible to migrate setuptools `setup.py` package build infrastructure to pep517 package build infrastructure](https://github.com/buildroot/buildroot/commit/4686fb32a053753f502e24c5cd980db5b71f6a2c) due to setuptools pep517 support effectively requiring `setup.py` build infrastructure itself for bootstrapping.

davidhewitt:
Thanks for the heads up! I'm in no rush to push forward with a 2.0 release, so potentially we can just wait for upstream before we consider a 2.0 in that case.

jameshilliard:
> * Remove `Binding.Exec` (just use `RustBin`).

I think for embedding binaries in site-packages [like in selenium](https://github.com/SeleniumHQ/selenium/blob/selenium-4.25.0/py/setup.py#L87-L92) we still may need to keep `Binding.Exec` as `RustBin` is only for stuff that goes in the scripts path AFAIU.

PR #272, opened 2022-07-27
Replace `setuptools-scm` with bumpversion

messense:
See the discussion in https://github.com/PyO3/setuptools-rust/pull/271.

* patch release: `bumpversion patch`
* minor release: `bumpversion minor`
* major release: `bumpversion major`

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,8 @@
+[bumpversion]
+commit = True
+tag = False
+current_version = 1.4.1
+
+[bumpversion:file:setup.cfg]
+
+[bumpversion:file:setuptools_rust/version.py]
--- a/setuptools_rust/version.py
+++ b/setuptools_rust/version.py
@@ -1,0 +1,4 @@
+__version__ = version = "1.4.1"
+__version_tuple__ = version_tuple = tuple(
+    map(lambda x: int(x[1]) if x[0] < 3 else x[1], enumerate(__version__.split(".")))
+)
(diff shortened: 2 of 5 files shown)


setuptools_rust/version.py
```diff
@@ -0,0 +1,2 @@
+__version__ = version = "1.4.1"
+__version_tuple__ = version_tuple = tuple(__version__.split("."))
```
messense, line 2: TODO: I think it needs to be converted to int.

setuptools_rust/version.py
```diff
@@ -0,0 +1,4 @@
+__version__ = version = "1.4.1"
+__version_tuple__ = version_tuple = tuple(
+    map(lambda x: int(x[1]) if x[0] < 3 else x[1], enumerate(__version__.split(".")))
```
messense, line 3: This `__version_tuple__` is for compatibility with previous `setuptools-scm`, it's kinda ugly, we could remove it in 2.0

.bumpversion.cfg
```diff
@@ -0,0 +1,8 @@
+[bumpversion]
+commit = True
+tag = True
```
davidhewitt, line 3: I wonder if we want `tag = False` ? Usually I let Github releases create the tag.

setuptools_rust/version.py
davidhewitt: 👍 let's note it in #259

.bumpversion.cfg
messense: OK

davidhewitt (approved):
Looks good to me, thanks!

PR #273, opened 2022-07-28
Fix RustBin build without wheel

jameshilliard:
We shouldn't depend on wheel unless building a .whl file.

Fixes: [#1028](https://github.com/PyO3/maturin/issues/1028)


messense (commented):
Thanks! Please also add a [changelog](https://github.com/PyO3/setuptools-rust/blob/main/CHANGELOG.md) entry.

jameshilliard:
> Please also add a [changelog](https://github.com/PyO3/setuptools-rust/blob/main/CHANGELOG.md) entry.

Added

messense (approved):

Issue #274, opened 2022-07-29
RustBin doesn't get installed to correct scripts path with setup.py install


jameshilliard:
It looks like setuptools doesn't correctly install RustBin correctly due to a [lack of entrypoints](https://github.com/PyO3/setuptools-rust/blob/v1.4.1/setuptools_rust/extension.py#L329-L330) or some issue with using a build path for the install path. Installation seems to work when using pep517 based installs but not normal setuptools installs(ie `python setup.py install`).

Note this issue occurs both with and without the [wheel path build fix](https://github.com/PyO3/setuptools-rust/commit/5be25fee1dec303e000a8a7b5a2b042ccd4c45ad) and regardless of whether or not the wheel dependency is installed.

The build path looks like this(not sure that's correct script build path when not using wheels):
```
build/lib.linux-x86_64-cpython-310/maturin-0.13.0.data/scripts/maturin
```

However the install path looks wrong:
```
venv/lib/python3.10/site-packages/maturin-0.13.0-py3.10-linux-x86_64.egg/maturin-0.13.0.data/scripts/maturin
```

If I build a wheel and install with [installer](https://github.com/pypa/installer) it does end up in the right place:
```
venv/bin/maturin
```

messense:
That's exactly what I meant it's a wheel specific thing before: https://github.com/PyO3/maturin/issues/1028#issuecomment-1197728346.

> not sure that's correct script build path when not using wheels

I don't know either.

jameshilliard:
> I don't know either.

Hmm, maybe we need to override something in [`install_scripts`](https://github.com/pypa/setuptools/blob/main/setuptools/command/install_scripts.py)?

PR #275, opened 2022-07-30
Fix RustBin setuptools install

jameshilliard:
Fixes: #274


messense (commented):
Thanks, I think we should add a test case to ensure that it does not regress in the future, right now we only test with wheel.

https://github.com/PyO3/setuptools-rust/blob/main/examples/hello-world/noxfile.py

jameshilliard:
> Thanks, I think we should add a test case to ensure that it does not regress in the future, right now we only test with wheel.

Yeah, think I got the setuptools only install working in a test, along with an install exclusion filter so that setuptools doesn't also install scripts in places that don't make sense like:
```
venv/lib/python3.10/site-packages/maturin-0.13.0-py3.10-linux-x86_64.egg/maturin-0.13.0.data/scripts/maturin
```

messense (approved):
Thanks, looks great!

davidhewitt:
👍 thanks!

PR #277, opened 2022-07-30
Simplify `install_rust_extension` and `bdist_wheel_rust_extension`

### LINKED ISSUE #276: Simplify `install_rust_extension` command
While reviewing #275 I noticed that the comment of [`install_rust_extension`](https://github.com/PyO3/setuptools-rust/blob/5be25fee1dec303e000a8a7b5a2b042ccd4c45ad/setuptools_rust/setuptools_ext.py#L154) command says:

> this is required because, install directly access distribution's ext_modules attr to check if dist has ext modules

That issue was [resolved in upstream](https://github.com/pypa/distutils/pull/43) long ago, let's try simplify or even remove `install_rust_extension` command override.

messense:
Closes #276


setuptools_rust/setuptools_ext.py
```diff
@@ -152,19 +152,10 @@ def run(self) -> None:
...
-            # all ext modules
-            mods = []
-            if self.distribution.ext_modules:
-                mods.extend(self.distribution.ext_modules)
             if self.distribution.rust_extensions:
-                mods.extend(self.distribution.rust_extensions)
-
                 scripts = []
                 for ext in self.distribution.rust_extensions:
                     scripts.extend(ext.entry_points())
```
messense, line 161: I wonder if we can get rid of this.

jameshilliard: Maybe needed by the [scripts/exec bindings](https://github.com/PyO3/setuptools-rust/blob/v1.4.1/setuptools_rust/extension.py#L216-L224) still?

messense: Yes, I'm thinking about whether it's possible to move it to `self.distribution` so that we don't need to override `finalize_options`.

jameshilliard: Probably a good idea to have a regular binding.exec+script style test similar to [the hello world one](https://github.com/PyO3/setuptools-rust/blob/main/examples/hello-world/setup.py).

examples/hello-world-script/setup.py
```diff
@@ -0,0 +1,18 @@
...
+from setuptools_rust import RustExtension, Binding
+
+setup(
+    name="hello-world-script",
+    version="1.0",
+    rust_extensions=[
+        RustExtension(
+            {"hello-world-script": "hello_world.hello-world-script"},
+            binding=Binding.Exec,
+            script=True,
```
davidhewitt, line 12: To check I understand correctly, we won't need `examples/hello-world-script` when we drop `Binding.Exec` in 2.0?

davidhewitt (approved):
Looks like a good simplification to me, thanks 👍

examples/hello-world-script/setup.py
messense: Yes.

PR #278, opened 2022-07-30
Test all examples with setuptools based install

jameshilliard:
Probably a good idea to test all the examples with setuptools based installs.


messense (approved):
Thanks!

PR #280, opened 2022-08-10
fix `get_lib_name`

tomtau:
before it fails with `TypeError: _metadata() missing 1 required positional argument: 'cargo'`


davidhewitt:
Thanks for opening this - I've pushed a changeling entry and a test to avoid this break again. Looks like `mypy` probably didn't catch this because of the `@lru_cache` decorator - that's unfortunate.

PR #291, opened 2022-09-19
packaging: fix missing `examples/` from sdist

davidhewitt:
Going to try installing `setuptools_scm` when building the sdist. Pushing a draft PR here to try it out.


davidhewitt:
Comparing the two dry runs it's clear that installing `setuptools-scm` is enough to ensure all git files are included in the sdist.

I'll force-push a real patch, merge & release.

Issue #257, opened 2022-06-29
Better error when compiled artifacts are in use


davidhewitt:
I was thinking the other day it might be nice to check if any processes are using the installed extension before attempting to copy across.

I think on Windows we'll get an `OsError` raised during copy, and on Unix any running processes are just likely to crash when they next need to load a segment of the native code.

Doing a check and raising our own friendly error might be nicer UX.

adamreichold:
> I think when I've seen segfaults before it was for libraries rather than executables. Might behave differently?

I think this mostly happens because updates are not transactional, e.g. an older executable will after having been started dynamically load a newer library that has an different ABI the older version of the library. But as written above, `ETXTBSY` was never completely reliable.

davidhewitt:
From the article you linked above:

> The MAP_DENYWRITE flag was created for just this use case, so that shared libraries could not be written while in use. When MAP_DENYWRITE went away, so did that protection; current Linux systems will happily allow a suitably privileged user to overwrite in-use, shared libraries. 

It really does sound to me like shared libraries don't get `ETXTBSY` protection 😊

adamreichold:
Indeed, the removal seems to be further along than I remembered from reading the article. Appears like I did not read it very thoroughly. 😊

adamreichold:
So for extensions modules which are shared libraries `ETXTBSY` is irrelevant. Sorry for the noise...

davidhewitt:
Not at all! Always worth sharing information and debating my spurious claims of segfaults 😂

PR #295, opened 2022-09-29
Fix segfaults caused by modifying existing shared library

erikjohnston:
When the built shared library is copied to the correct location, it actually *modifies* any existing shared library, which causes any running process using it to segfault.

To fix this, we first delete any existing shared library (which is safe to do) and then copy the newly built shared library to the correct location as a new file.

Related to #257

At least, this is the behaviour on Linux, I'm not sure about other OSes.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -346,26 +346,17 @@ class build_rust(RustCommand):
 
             log.info("Copying rust artifact from %s to %s", dylib_path, ext_path)
 
-            # We delete any existing library file before copying to avoid
-            # causing running processes that use it from segfaulting.
+            # We want to atomically replace any existing library file. We can't
+            # just copy the new library directly on top of the old one as that
+            # causes the existing library to be modified (rather the replaced).
+            # This means that any process that currently uses the shared library
+            # will see it modified and likely segfault.
             #
-            # This is because shared libraries are memory mapped into processes
-            # that use it, so modifying the shared library file (which is what
-            # copying the file does) means the next time a process calls into
-            # the shared library they get the updated library rather than the
-            # original one, causing confusion and segfaults.
-            #
-            # Deleting a memory mapped file, on the other hand, is correctly
-            # handled by the OS, i.e. a copy is kept around until the running
-            # process exits. Thus, deleting the existing library file and *then*
-            # copying the new one won't change the existing memory mapped file,
-            # and so is safe to do.
-            try:
-                os.remove(ext_path)
-            except FileNotFoundError:
-                pass
-
-            os.rename(dylib_path, ext_path)
+            # We first copy the file to the same directory, as `os.rename`
+            # doesn't work across file system boundaries.
+            temp_ext_path = ext_path + "~"
+            shutil.copyfile(dylib_path, temp_ext_path)
+            os.rename(temp_ext_path, ext_path)
 
             if sys.platform != "win32" and not debug_build:
                 args = []
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -356,7 +356,7 @@ class build_rust(RustCommand):
             # doesn't work across file system boundaries.
             temp_ext_path = ext_path + "~"
             shutil.copyfile(dylib_path, temp_ext_path)
-            os.rename(temp_ext_path, ext_path)
+            os.replace(temp_ext_path, ext_path)
 
             if sys.platform != "win32" and not debug_build:
                 args = []
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -345,6 +345,26 @@ class build_rust(RustCommand):
                 os.makedirs(os.path.dirname(ext_path), exist_ok=True)
 
             log.info("Copying rust artifact from %s to %s", dylib_path, ext_path)
+
+            # We delete any existing library file before copying to avoid
+            # causing running processes that use it from segfaulting.
+            #
+            # This is because shared libraries are memory mapped into processes
+            # that use it, so modifying the shared library file (which is what
+            # `shutil.copyfile` does) means the next time a process calls into
+            # the shared library they get the updated library rather than the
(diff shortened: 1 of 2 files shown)


setuptools_rust/build.py
```diff
@@ -345,6 +345,26 @@ def install_extension(
...
+            # handled by the OS, i.e. a copy is kept around until the running
+            # process exits. Thus, deleting the existing library file and *then*
+            # copying the new one won't change the existing memory mapped file,
+            # and so is safe to do.
+            try:
+                os.remove(ext_path)
+            except FileNotFoundError:
+                pass
+
             shutil.copyfile(dylib_path, ext_path)
```
adamreichold, line 368: If we try to make this more robust, wouldn't an atomic replacement be better here, e.g. move to `".~" + ext_path` and then move `".~" + ext_path` to `ext_path` to atomically replace the name with the copy of the library? Otherwise it seems possible that processes that try to start and map a partially copied library might fail as well.

erikjohnston: Yeah, we could do this as a `os.rename(..)` after deleting the existing file. I don't follow why we would do two moves?

adamreichold: As it is implemented now, there is a window after removing the existing library and until moving the new one in its place when there is no library there. This would be avoid by the copy-then-rename dance.

(I am also not sure whether moving the library is the right thing to do since it removes the build output. `os.rename` also cannot cross file system boundaries which is why copying the file into the target directory first is usually preferable.)

erikjohnston: > As it is implemented now, there is a window after removing the existing library and until moving the new one in its place when there is no library there. This would be avoid by the copy-then-rename dance.

I didn't think there was a way to atomically swap two files? 

>  (I am also not sure whether moving the library is the right thing to do since it removes the build output. os.rename also cannot cross file system boundaries which is why copying the file into the target directory first is usually preferable.)

Aha! Right, that makes sense, so the `shutil.copyfile` should become `shutil.copyfile` to same directory then a `rename` to the actual path? Though I think we still need to delete the existing library before the rename, as rename fails if the path already exists?

adamreichold: > Though I think we still need to delete the existing library before the rename, as rename fails if the path already exists?

I am not totally sure on Windows, but on Unix-like systems this exactly the idiom to atomically replace the file. (It is not guaranteed by POSIX IIRC but basically all real Unix filesystems guarantee in their common configurations.)

adamreichold: (Meaning one can use `rename` to overwrite a file without removing it first.)

adamreichold: Just had a look at the Python docs and it does not work on Windows but it is required by POSIX nowadays: https://docs.python.org/3/library/os.html#os.rename

But they also recommend `os.replace` as the cross platform way of achieving exactly this: https://docs.python.org/3/library/os.html#os.replace

So I think copying to the target directory and then `os.replacing` without removing first would be the best approach.

erikjohnston: Huh, well TIL. Thank you very much! 

I don't know what I was reading when I was previously looking at the docs, but [`os.rename`](https://docs.python.org/3/library/os.html#os.rename) clearly states what's going on:

> On Windows, if dst exists a [FileExistsError](https://docs.python.org/3/library/exceptions.html#FileExistsError) is always raised.
>
> On Unix, if both are files, dst will be replaced silently if the user has permission. The operation may fail on some Unix flavors if src and dst are on different filesystems. If successful, the renaming will be an atomic operation (this is a POSIX requirement).

So I think on Windows we'd still need to do the `os.remove`? But equally, I'm not sure if Windows suffers the same problem, so we could just leave it as is?

setuptools_rust/build.py
```diff
@@ -345,7 +345,18 @@ def install_extension(
...
+
+            # We want to atomically replace any existing library file. We can't
+            # just copy the new library directly on top of the old one as that
+            # causes the existing library to be modified (rather the replaced).
+            # This means that any process that currently uses the shared library
+            # will see it modified and likely segfault.
+            #
+            # We first copy the file to the same directory, as `os.rename`
+            # doesn't work across file system boundaries.
+            temp_ext_path = ext_path + "~"
```
erikjohnston, line 357: It felt easier to just prefix the `ext_path` with `~`, happy to name it something else if folk think that's worthwhile.

adamreichold: Yeah, the actual pattern is pretty inconsequential as far I can see. So this is certainly fine with me.

setuptools_rust/build.py
```diff
@@ -345,7 +345,18 @@ def install_extension(
...
+            shutil.copyfile(dylib_path, temp_ext_path)
+            os.rename(temp_ext_path, ext_path)
```
adamreichold, line 359: Shouldn't this be

```suggestion
            os.replace(temp_ext_path, ext_path)
```

due to the Windows part?

adamreichold: (I think the CI actually hits this...)

erikjohnston: Yes :facepalm:

adamreichold (approved):

erikjohnston:
Thank you for the review! :heart:

davidhewitt (commented):
Thanks, logic looks good to me, could you just add a CHANGELOG entry please? Then let's merge 👍

erikjohnston:
I've added some words to the changelog, though I don't think I really expressed it well.

davidhewitt (approved):
That's great as-is - thanks! 👍

richvdh:
@davidhewitt: any chance of a release to include this change? It caused a significant outage for us today :(.

davidhewitt:
Sorry for the delay on release here. I'll try to put live tomorrow.

davidhewitt:
See #331

erikjohnston:
Thank you! :heart:

PR #303, opened 2022-11-16
CI: Add `abi3audit`

messense:
https://blog.trailofbits.com/2022/11/15/python-wheels-abi-abi3audit/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,7 @@ jobs:
       - name: Build an abi3 wheel
         shell: bash
         run: |
+          set -e
           cd examples/rust_with_cffi/
           python --version
           pip install -U wheel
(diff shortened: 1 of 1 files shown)


adamreichold:
If I understand things correctly, any errors indicated here would most likely stem from incorrect FFI definitions shipped with PyO3 (or less likely setuptools-rust not passing the feature flags to PyO3) but we have to test it here since it works against full wheels (so either here or over at maturin)?.

So to make this more useful, maybe we could run this using a Git dependency on the PyO3's current `main` branch and on a schedule similar to https://github.com/PyO3/pyo3-ffi-check?

messense:
https://github.com/PyO3/setuptools-rust/blob/8cc148169c3814969e0f9fb184db707174240562/.github/workflows/ci.yml#L166

We test it here because setuptools requires `--py-limited-api`, maturin only relies on these `abi3-*` Cargo features.

IMO we should run `pyo3-ffi-check` on every PR that changes `pyo3-ffi` not just on a schedule.

adamreichold:
> We test it here because setuptools requires --py-limited-api, maturin only relies on these abi3-* Cargo features.

Understand. But both could create incorrectly tagged abi3 extensions due to either errors in the `pyo3-ffi` crate or by messing up how those feature flags are passed to Cargo, couldn't they? So Maturin could have a similar step in its CI to make sure it does not?

> IMO we should run pyo3-ffi-check on every PR that changes pyo3-ffi not just on a schedule.

Agreed that this would be even better. (Maybe we could tie this change in with the move to bors over at PyO3?)

But it seems to orthogonal to making the abi3audit check work against PyO3's `main` branch in the first place?

messense:
> Maybe we could tie this change in with the move to bors over at PyO3?

I think it's a good idea.



> But it seems to orthogonal to making the abi3audit check work against PyO3's `main` branch in the first place?

I'll work on that.

davidhewitt:
> IMO we should run pyo3-ffi-check on every PR that changes pyo3-ffi not just on a schedule.

Agreed with this, so far I have massively failed to configure Bors 😂

PR #321, opened 2023-03-29
Fix compilation bug in toolchain >= 1.70.0

### LINKED ISSUE #320: Compilation in setuptools_rust fails with `1.70.0-nightly` toolchain
# Assumptions

- problematic toolchain
-- `rustc 1.70.0-nightly (db0cbc48d 2023-03-26)`
-- `rustc 1.70.0-nightly (478cbb42b 2023-03-28)`
- toolchain not affected
-- `rustc 1.68.0-stable`

# Description

`wheel` package generation fails with latest toolchain, when `crate-type = ["cdylib"]` is not contained in `Cargo.toml`.

# Cause

The following command in build process will not work:

```
cargo rustc --lib --message-format=json-render-diagnostics --manifest-path Cargo.toml --release -v --features 'python pyo3/extension-module' --no-default-features -- --crate-type cdylib
```

Instead, following works:

```
cargo rustc --lib --message-format=json-render-diagnostics --manifest-path Cargo.toml --release -v --features 'python pyo3/extension-module' --no-default-features --crate-type cdylib
```

# Workaround

any of the following avoids the bug:

- Write `crate-type = ["cdylib"]` in `Cargo.toml`
- Write `args=["--crate-type", "cdylib"]` in initializer of `RustExtension`

Sample:

```py
                RustExtension(
                    'namespace.package', 'Cargo.toml',
                    binding = Binding.PyO3,
                    args = ["--crate-type", "cdylib"],
                )
```

I will create PRs later.

yasuo-ozu:
Fix #320

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -28,7 +28,8 @@ from typing_extensions import Literal
 from ._utils import format_called_process_error
 from .command import RustCommand
 from .extension import Binding, RustBin, RustExtension, Strip
-from .rustc_info import get_rust_host, get_rust_target_list, get_rustc_cfgs
+from .rustc_info import get_rust_host, get_rust_target_list, get_rustc_cfgs, get_rust_version
+from semantic_version import Version
 
 
 class build_rust(RustCommand):
@@ -142,6 +143,7 @@ class build_rust(RustCommand):
 
         quiet = self.qbuild or ext.quiet
         debug = self._is_debug_build(ext)
+        is_new_toolchain = get_rust_version() >= Version('1.70.0-nightly')
 
         cargo_args = self._cargo_args(
             ext=ext, target_triple=target_triple, release=not debug, quiet=quiet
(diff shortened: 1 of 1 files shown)


adamreichold:
Is this behaviour on a path towards stabilization? Should we already commit to this approach?

setuptools_rust/build.py
```diff
@@ -142,6 +143,7 @@ def build_extension(
 
         quiet = self.qbuild or ext.quiet
         debug = self._is_debug_build(ext)
+        is_new_toolchain = get_rust_version() >= Version('1.70.0-nightly')
```
adamreichold, line 146: I think a more description name would be helpful (today's new is tomorrow's old). Also do we need to include "nightly" here? Can we compare directly against "1.70.0" or this considered larger than "1.70.0-nightly" by `Version`?

yasuo-ozu: Thank you. I will withdraw this PR and fix it in another PR.

yasuo-ozu: I wrote `Version('1.70.0-nightly')` here, because I want `is_new_toolchain` to be `True` when `get_rust_version() == Version("1.70.0-nightly")`.
But in fact `Version('1.70.0-nightly') >= Version('1.70.0')` is `False`.

Do you think something like the following is better?

```py
rustc_version = get_rust_version()
is_new_toolchain = rustc_version.major > 1 or (rustc_version.major == 1 and rustc_version.minor >= 70)
```

adamreichold: Yes, I think that would be preferable. (Especially if combined with a more descriptive name than "new".)

PR #322, opened 2023-03-29
Fix compilation bug in toolchain >= 1.70.0

### LINKED ISSUE #320: Compilation in setuptools_rust fails with `1.70.0-nightly` toolchain

yasuo-ozu:
Fix #320

Please see the discussion in https://github.com/PyO3/setuptools-rust/pull/321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Fixed
 - Fix a bug where rebuilding the library would cause any running processes using it to segfault. [#295](https://github.com/PyO3/setuptools-rust/pull/295)
 - Fix `setup.cfg` format for compatibility with "poetry==1.4.0". [#319](https://github.com/PyO3/setuptools-rust/pull/319)
+- Fix to use `--crate-type` option of cargo if "toolchain >= 1.64". [#322](https://github.com/PyO3/setuptools-rust/pull/322)
 
 ## 1.5.2 (2022-09-19)
 ### Fixed
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -212,7 +212,7 @@ class build_rust(RustCommand):
             ):
                 rustc_args.extend(["-C", f"link-args=-sSIDE_MODULE=2 -sWASM_BIGINT"])
 
-            if is_toolchain_1_70_or_later:
+            if is_toolchain_1_70_or_later and "--crate-type" not in cargo_args:
                 cargo_args.extend(["--crate-type", "cdylib"])
 
             command = [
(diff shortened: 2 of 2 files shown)


yasuo-ozu:
I think the mypy error is occured out of my code, but I don't know why it fails while the `main` branch is passed.

davidhewitt:
Are you sure this is an intentional change? I don't see it mentioned in https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md

yasuo-ozu:
Actually I cannot figure out which change on rustc break the compatibility.
(it occurs between ff4b772 ... 8a73f50)

Recently --crate-type option is implemented in cargo, so I think it is safer to use --crate-type option on cargo.

yasuo-ozu:
@davidhewitt I think the mypy error is not related on this PR, so is it good for you to fix it in the same branch?

davidhewitt:
Yes, I can fix the mypy error separately.

Argument for using cargo options seems reasonable to me. It looks like `--crate-type` was added in cargo 1.64, so can you please rework this PR to use the new options for 1.64 or later?

We can then document this in the CHANGELOG as "changed" rather than "fixed".

yasuo-ozu:
Thanks a lot. I fixed the version and modified CHANGELOG.
Could you please release it after it is merged?

davidhewitt (approved):
Thanks - I'll try to tidy things up on `main` with a view to releasing tomorrow.

PR #329, opened 2023-04-24
Remove Python 3.6 trove classifier & add 3.11

smheidrich:
Support for Python 3.6 was dropped in https://github.com/PyO3/setuptools-rust/commit/49e9000c661ece1d8b4a6a56479aea760bf5516d but the Trove classifiers still list it as a compatible version. This PR fixes that.


adamreichold:
We do support and test 3.11 though. Could you add this here as well? Thank you.

smheidrich:
Sure, done

adamreichold:
mypy error is unrelated. Merging.

PR #336, opened 2023-05-13
chore(distutils): remove distutils usages

stegayet:
## Description

[`distutils` module usage is deprecated](https://peps.python.org/pep-0632/) since Python 3.10 and will be removed in Python 3.12, this PR gets rid of remaining usages of this module.

Replacements are taken from https://peps.python.org/pep-0632/#migration-advice and https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html.

## How this has been tested ?
To be checked

### DIFF
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -1,11 +1,12 @@
 import os
 import subprocess
 import sysconfig
-from distutils import log
-from distutils.command.clean import clean
+import logging
+
 from typing import List, Set, Tuple, Type, cast
 
 from setuptools.command.build_ext import build_ext
+
 from setuptools.command.install import install
 from setuptools.command.install_lib import install_lib
 from setuptools.command.install_scripts import install_scripts
@@ -19,6 +20,8 @@ try:
     from wheel.bdist_wheel import bdist_wheel
 except ImportError:
     bdist_wheel = None
+
+logger = logging.getLogger(__name__)
 
 
 def add_rust_extension(dist: Distribution) -> None:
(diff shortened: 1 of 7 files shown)


stegayet:
`from distutils.command.clean import clean` is discontinued and has no replacement solution as per https://github.com/pypa/setuptools/discussions/2838, I need to dig into this issue.

setup.cfg
```diff
@@ -35,8 +35,8 @@ setup_requires = setuptools>=62.4
 python_requires = >=3.7
 
 [options.entry_points]
-distutils.commands =
+setuptools.commands = 
     clean_rust=setuptools_rust:clean_rust
     build_rust=setuptools_rust:build_rust
-distutils.setup_keywords =
+setuptools.setup_keywords =
```
davidhewitt, line 41: It looks to me like these have not changed; I suggest reverting these changes (or alternatively raising PRs to setuptools to rename these).

setuptools_rust/setuptools_ext.py
```diff
@@ -1,8 +1,8 @@
 import os
 import subprocess
 import sysconfig
-from distutils import log
-from distutils.command.clean import clean
+import logging
+from setuptools.command.clean import clean
```
davidhewitt, line 5: This is only used for type checking, so I think you could leave it as distutils and put it under an `if TYPE_CHECKING` guard?

davidhewitt (commented):
Thanks for working on this! Some initial suggestions which may help out...

setup.cfg
stegayet: Good catch :+1:

setuptools_rust/setuptools_ext.py
stegayet: It seems to be used not only for typing as it breaks the CI, trying to use vendorized version of `distutils` in the meantime.

davidhewitt:
@stegayet I've merged #337 which fixed a lot of CI issues, you may find it helpful to rebase.

davidhewitt:
For now I've rebased and removed the import of `clean` (just going to duck-type it).

stegayet:
Did not have time to keep this work going further, sorry for that :+1:

PR #337, opened 2023-05-15
migrate to pyproject.toml

davidhewitt:
This seemed like a worthwhile cleanup...


pyproject.toml
```diff
@@ -1,3 +1,48 @@
+[project]
+name = "setuptools-rust"
+version = "1.6.0"
+description = "Setuptools Rust extension plugin"
+readme = "README.md"
+requires-python = ">=3.7"
+license = {file = "LICENSE"}
```
adamreichold, line 7: I think it would be helpful if this used some machine readable form like an SPDX identifier? Or is the MIT license used customized?

adamreichold (approved):

pyproject.toml
davidhewitt: I think it's standard MIT; we have a trove classifier which documents it as MIT License.

I got the impression from https://peps.python.org/pep-0621/#license that I should not use an SPDX identifier here? @messense maybe you know what's the best way to set the license as MIT?

davidhewitt:
I think I prematurely pushed this, looks like CI is bust in a few places. I'll fixup tomorrow.

pyproject.toml
adamreichold: I think the current version of this information is https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#license together with https://packaging.python.org/en/latest/specifications/core-metadata/#license which indeed does not suggest something like SPDX identifier.

My reading of the second link is actually that the `license` field should be left out completely if the license trove classifier fully captures the license.

davidhewitt: > My reading of the second link is actually that the license field should be left out completely if the license trove classifier fully captures the license.

Yes, I had thought the same though it wasn't perfectly clear. I've dropped `license` for now.

davidhewitt:
CI is green :tada:

PR #343, opened 2023-06-25
Pass include_dirs to cargo using env

jameshilliard:
Sometimes the cargo build script needs to know the `include_dirs`, lets pass this in the `SETUPTOOLS_RUST_INCLUDE_DIRS` env variable.

See [#9132](https://github.com/pyca/cryptography/pull/9132) for example usage.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -81,6 +81,7 @@ class build_rust(RustCommand):
         self.build_temp = None
         self.plat_name = None
         self.build_number = None
+        self.include_dirs = None
         self.target = os.getenv("CARGO_BUILD_TARGET")
         self.cargo = os.getenv("CARGO", "cargo")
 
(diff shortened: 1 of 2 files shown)


davidhewitt:
How do these get passed to C / C++ compilations? I wonder if we can expose them in the same way and then let downstream projects handle them in the same way as if they were a C/C++ compilation. This might lead to best automatic support

(e.g. should we maybe set `CFLAGS`?)

jameshilliard:
> How do these get passed to C / C++ compilations?

I think they are just [passed](https://github.com/pypa/setuptools/blob/v68.0.0/setuptools/_distutils/ccompiler.py#L1212-L1213) to the compiler command line.

> I wonder if we can expose them in the same way and then let downstream projects handle them in the same way as if they were a C/C++ compilation.

I mean, that's more or less what I'm trying to do here. The cargo build script in my cryptography pull request passes the include_dirs from the env variable to the compiler using `-I` command line include arguments(by setting them using `build.include(include_path)`).

> (e.g. should we maybe set CFLAGS?)

This seems like it could potentially cause conflicts if CFLAGS are being set in the env already.

alex:
FWIW, we ended up solving this with https://github.com/pyca/cryptography/pull/9129

jameshilliard:
I'll close this for now then.

Issue #208, opened 2022-02-26
Ability to specify rust_extensions via setup.cfg


alex:
Currently one specifies `rust_extensions` in `setup.py`, imperatively (e.g. https://github.com/pyca/cryptography/blob/main/setup.py#L43-L56). Nowadays, setuptools lets one specify many options declaratively in `setup.cfg` (e.g. https://github.com/pyca/cryptography/blob/main/setup.cfg). It'd be great if `rust_extensions` could also be specified in `setup.cfg` -- though truthfully I have no idea if that is possible!

davidhewitt:
I agree that this would be great. I think that this needs support from `setuptools` maintainers. As far as I know, [this page](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) describes all the valid `setup.cfg` fields and I'm not aware of a mechanism for plugins like `setuptools_rust` to define additional fields.

messense:
FYI: a related upstream issue: https://github.com/pypa/setuptools/issues/1272

messense:
Although we can not extend `setup.cfg`, we can certainly extend `pyproject.toml` to add a `[tool.setuptools-rust]` section, for example

```toml
[[tool.setuptools-rust.rust-extensions]]
target = "example"
binding = "pyo3"
```

Then we can implement a PEP 517 backend in setuptools-rust to read it and generate a corresponding `setup.py`.

```toml
[build-system]
requires = ["setuptools", "wheel", "setuptools-rust"]
build-backend = "setuptools_rust"
```

alex:
At least for my use case I need the build backend to be setuptools itself, since I also have cffi extensions that need _its_ setuptools extension.

I don't we can do configuration in `pyproject.toml` without overriding `build-backend`?

messense:
Right, but it should work after merging https://github.com/pyca/cryptography/pull/7164

alex:
Good point! Though that's still a ways off since it needs patches to rust and pypy.

messense:
Or we can add a `[tool.setuptools-rust.extra-setup-args]` that can add arbitrary arguments to the generated `setup.py`.

```toml
[tool.setuptools-rust.extra-setup-args]
cffi_modules = ["src/_cffi_src/build_openssl.py:ffi"]
```

nils-werner:
CFFI has implemented the `cffi_modules` keyword for `setup(...)`, which lets you circumvent the issue of having to import `cffi` at install time:

```python
[...]

PR #348, opened 2023-08-03
Allow users to use `pyproject.toml` define `RustExtension` and `RustBin` (instead of `setup.py`)

abravalheri:
This PR was created with the discussion of #208 in mind.

It allows users to specify Rust extensions in the same way they would using `setup.py`, but using `pyproject.toml` instead.
For an example of how this would work, see `examples/hello-world-pyprojecttoml/pyproject.toml`.

Approach:

- `setuptools` exposes a [`setuptools.finalize_distribution_options` hook](https://setuptools.pypa.io/en/latest/userguide/extension.html#customizing-distribution-options) that allows reusing the logic implemented in `setuptools_rust.setuptools_ext.rust_extensions` for modifying the `dist` object based on the `pyproject.toml` config.
- What is implemented in this PR is pretty much a 1:1 translation between the way `setup.py` is currently used by `setuptools_rust` to `pyproject.toml`
  - I suppose it could be done in a smarter way (e.g. one could even try to use `Cargo.toml` info to automatically fill in the fields in the `dist` object, even things like `name` or `version`), but this is the most obvious/easier approach, so I just went with it.
- The kwargs for the `RustExtension` and `RustBin` are translated directly to `pyproject.toml`, the only differences are:
  - `pyproject.toml` uses `-` instead of `_` (by convention, see PEP 517/621)
  - `enum` objects are represented in a "stringfied manner", e.g. `Binding.PyO3 => "PyO3"`.
- To differentiate between `RustExtension` and `RustBin`, the user can specify `[[tool.setuptools-rust.ext-modules]]` and `[[tool.setuptools-rust.binaries]]`.
  - I went with these names because they were the most obvious in my opinion but please feel free to change if you prefer something different. 
- The example I added follows what I would implement in a project of mine (a private module implemented in Rust + a layer of Python wrapping things). It also makes it easier.

Remarks:

- I am not a Rust developer, so my example can be ugly. Please feel free to change it.
- If the project would like to use a different approach for `pyproject.toml`, please feel free to close this PR.
- Please feel free to modify the PR directly if you would like to tweak things around (e.g. any form of bike-shedding)


examples/hello-world-pyprojecttoml/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[profile.release-lto]
+inherits = "release"
+lto = true
+
+[lib]
+# See https://github.com/PyO3/pyo3 for details
+name = "_lib"  # private module to be nested into Python package
+crate-type = ["cdylib"]
+path = "rust/lib.rs"
```
davidhewitt, line 19: Personally I would prefer if we left the rust directory as `src`, because that's what the rest of the examples do. I think there's no good reason to diverge here, but I'm also happy to be persuaded that `rust` is better. If `rust` is agreed to be better, we should update the rest of the examples to use it too.

examples/hello-world-pyprojecttoml/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[tool.setuptools.packages]  # pure Python packages/modules
+find = { where = ["python"] }
+
+[[tool.setuptools-rust.ext-modules]]
+target = "hello_world._lib"  # private module to be nested into Python package
+py-limited-api = "auto"
+args = ["--profile", "release-lto"]
+
+[[tool.setuptools-rust.binaries]]  # executable to be installed in `$venv/bin`
```
davidhewitt, line 20: I wonder if this should be `[[tool.setuptools-rust.bin]]` to match `Cargo.toml`'s `[[bin]]`?

examples/hello-world-pyprojecttoml/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "hello-world"
+version = "1.0"
+
+[project.scripts]  # Python entry-point wrapper to be installed in `$venv/bin`
+sum-cli = "hello_world.sum_cli:main"
+
+[tool.setuptools.packages]  # pure Python packages/modules
+find = { where = ["python"] }
+
+[[tool.setuptools-rust.ext-modules]]
```
davidhewitt, line 15: Similar to `[[tool.setuptools-rust.bin]]` suggestion below, should this be `[[tool.setuptools-rust.lib]]`? I'm less sure about this one, mostly because `Cargo.toml` only allows one `[lib]` but we allow multiple here.

examples/hello-world-pyprojecttoml/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[build-system]
+requires = ["setuptools", "wheel", "setuptools-rust"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "hello-world"
+version = "1.0"
+
+[project.scripts]  # Python entry-point wrapper to be installed in `$venv/bin`
+sum-cli = "hello_world.sum_cli:main"
```
davidhewitt, line 10: It might be a nice example to move this to be a Rust `#[pyfunction]` as the entry point, similar to https://www.maturin.rs/bindings#both-binary-and-library

davidhewitt (commented):
Nice, this is delightfully simple and a greatly appreciated solution to something which I simply hadn't had time to look into.

Given that `setup.py` configuration is sort of deprecated, how do you feel about going further and migrating all examples (except perhaps for a `hello-world-setuppy` to test the `setup.py` form)?

We should also update documentation to cover (recommend?) this layout. Again, I'd be tempted to move the documentation of using `setup.py` to a secondary page in the docs and make this `pyproject.toml` form the preferred option.

examples/hello-world-pyprojecttoml/Cargo.toml
davidhewitt: (See also https://github.com/PyO3/setuptools-rust/pull/350 - I've proposed moving to use `python/` subdirectory in all examples.)

abravalheri: Hi @davidhewitt, thank you very much for the review.

As a very personal approach, I would avoid as much as I can to re-use the `src` directory for storing the Rust files. The reason behind this is that both Python/setuptools and Rust kind of compete for the convention of using the `src` directory. For example, when the user does not specify a `packages = [...]` configuration in `pyproject.toml`, `setup.cfg` and `setup.py`, setuptools might attempt to perform [automatic discovering](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html), which will prioritize `src` as the convention for holding Python files meant for distribution.

Please note that the convention for using the `src` folder in Python is very old and well known in the community (e.g. [this blog post](https://blog.ionelmc.ro/2014/05/25/python-packaging/)). It even figures out in the [Packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/) and [Python packaging website](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/).

Since both languages and tools (setuptools and Cargo) have a convention for using the `src` folder, I preferred to side step the discussion and be explicit when specifying path.

From the point of view of a Python user that does not have experience with Rust development, I believe that an example showing things explicit is very useful and helps to reduce the learning overhead.

If that is OK with you, I would refrain for modifying this PR in that direction (since is a direction that I have personal problems with), but if you or other maintainers would like to go ahead and implement modifications in a follow up PR, I completely respect the project's motivations for doing so.

examples/hello-world-pyprojecttoml/pyproject.toml
abravalheri: I don't have any strong feelings about this change.
Can we have an agreement on all the changes in terms so I can implement all of them in batch?

examples/hello-world-pyprojecttoml/pyproject.toml
abravalheri: I don't have strong feelings about this one either.
However I do recognise that from the Python user perspective (e.g. a Python dev that finds himself in the situation where he has/wants to include Rust binaries in a package - such as wrapping Rust libraries or improving performance), `ext-modules` may be a more descriptive term.

On top of that, as you mentioned, Rust only seem to accept one `[lib]` (I don't have much Rust experience, so that came as a surprise for me yesterday - but I suppose you can side-step that by having multiple `Cargo.toml` files).

Once you have a final view on what you would like the terms to be, please let me know and I will implement a new commit.

examples/hello-world-pyprojecttoml/pyproject.toml
abravalheri: Thank you very much @davidhewitt, I can address this in a follow up commit that will also try to tackle the CI linter errors.

Would it be OK to add if I add a second entry-point. I think it is nice to show that you can mix and match Python and Rust modules...

abravalheri:
Hi @davidhewitt, thank you very much for the review.

> Given that setup.py configuration is sort of deprecated, how do you feel about going further and migrating all examples (except perhaps for a hello-world-setuppy to test the setup.py form)?

> We should also update documentation to cover (recommend?) this layout. Again, I'd be tempted to move the documentation of using setup.py to a secondary page in the docs and make this pyproject.toml form the preferred option.

I am happy to help with that once we finalize the round of reviews for the implementation. Maybe 2 follow up PRs? What do you think?

abravalheri:
> Given that setup.py configuration is sort of deprecated, how do you feel about going further and migrating all examples (except perhaps for a hello-world-setuppy to test the setup.py form)?

I would be interested in leaving at least 1 example (`hello-world-setuppy` as you suggested) using `setup.py`, because `setup.py` is still a perfectly valid configuration file, but deprecated as a primary CLI tool. (I think a nice analogy is that `setup.py` is becoming a configuration file in the style of `conftest.py` or `noxfile.py`: you don't run them directly in the terminal, but they are used under the hood by the tools they are related to).

examples/hello-world-pyprojecttoml/Cargo.toml
davidhewitt: Fair enough, let's keep `rust` directory here 👍

examples/hello-world-pyprojecttoml/pyproject.toml
davidhewitt: Sure, having both a Python entry-point and a Rust entry point suits me 👍

examples/hello-world-pyprojecttoml/pyproject.toml
davidhewitt: I think most the Python `pyproject.toml` metadata uses plural, so let's keep `ext-modules`.

examples/hello-world-pyprojecttoml/pyproject.toml
davidhewitt: I think `bin` is better than `binary` to match cargo, however I think given the comment below about plurals, let's go for `bins` perhaps? What do you think of that?

davidhewitt:
Thanks, I think just `binaries` -> `bins` and then this is good to merge. Follow-up PRs are fine with me and much appreciated 👍

davidhewitt:
(Also looks like `black` wants to format some more files :) )

examples/hello-world-pyprojecttoml/pyproject.toml
abravalheri: Done in https://github.com/PyO3/setuptools-rust/pull/348/commits/103993f2370b5cf25725a3a8f603bc8476509cf4

examples/hello-world-pyprojecttoml/Cargo.toml
abravalheri: Thank you very much ❤️ 💚 💙 💜

examples/hello-world-pyprojecttoml/pyproject.toml
abravalheri: No problems.

I implemented it in 1f58338d7134b93efe9bd9ff3e9dbd2e464dd726.
(`bins` might sound a bit weird, but I think that the argument you made about consistency is important).

davidhewitt:
Thanks @abravalheri, this looks good to me. I will rebase this and fix the merge conflict now. The macOS build failure is probably pre-existing, I'll merge this and see if I can work out a way to solve that separately to unblock you from any follow-up PRs you want to do.

messense:
> The macOS build failure is probably pre-existing

I think the linking error is the same as https://github.com/PyO3/maturin/issues/1080.

There is an implicit dependency from a binary to the library, see also https://github.com/rust-lang/cargo/issues/9235.

abravalheri:
Thank you very much @davidhewitt. I oversaw this one :P

examples/hello-world-pyprojecttoml/rust/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+
+/// Formats the sum of two numbers as string.
+#[pyfunction]
+fn sum_as_string(a: usize, b: usize) -> PyResult<String> {
+    Ok((a + b).to_string())
+}
+
+/// Calls Python (see https://pyo3.rs for details)
+#[pyfunction]
+fn demo(py: Python) -> PyResult<()> {
```
abravalheri, line 12: There seems to be a warning here because `py` is unused. Sorry I am not experienced in `Rust`, so I got confused by the following `Python::with_gil(|py| ...` statement.

```suggestion
fn demo(_py: Python) -> PyResult<()> {
```

If a maintainer would like to commit this to the PR, that would trigger the CI. Otherwise I can add it myself and wait for someone to trigger the CI.

davidhewitt: No problem, I've force-pushed again. (Removed `Python::with_gil` bit, as we can use `py` directly from the argument.)

abravalheri: Ah! Perfect, thank you @davidhewitt.

Still baby steps on Rust :P

davidhewitt:
Ok great, so this is all happy now except for the macOS failure this uncovered. I'll merge this now and separately I'll look into a solution for the macOS issue. Thanks @abravalheri

davidhewitt:
macOS fixed in #351

PR #351, opened 2023-08-11
set macos link args to fix bin + lib combination

davidhewitt:
Adding a `build.rs` with `pyo3_build_config::add_extension_module_link_args()` should fix the macOS failures in the new `pyproject.toml` example.

I had a look if there was a way for us to do this directly in `setuptools_rust`, but I couldn't see an obvious solution.


davidhewitt:
cc @messense - if this works then it's presumably a workaround for https://github.com/PyO3/maturin/issues/1080

davidhewitt:
Success!

PR #352, opened 2023-08-11
Change examples in accordance to review comments in #348

abravalheri:
This PR attempts to target a few review comments in #348, specifically:

---

#### Request to change examples from `setup.py` to `pyproject.toml`:

https://github.com/PyO3/setuptools-rust/pull/348#pullrequestreview-1562477607
> Given that setup.py configuration is sort of deprecated, how do you feel about going further and migrating all examples (except perhaps for a hello-world-setuppy to test the setup.py form)?

I tried to implement this from commits c0034a944f011de253acd6c0dde746b93fe8ccb8 to e00ab3099d2d7cf7fae5245ab9f602ba082ba550.

However there was a bunch of stuff that was difficult to understand because I don't have experience with Rust  or `setuptools-rust`, so while I was implementing the change I also added a bunch of comments kind of explaining for myself what was going on. I am not 100% sure if the comments are correct, or if maintainers agree with the comments, or if the maintainers think the comments are desirable in the examples. So please let me know and that can be changed.

A few other remarks:

- I renamed `hello-world-pyprojecttoml` to `hello-world` and `hello-world` to `hello-world-setuppy`.

- I removed direct calls to `python setup.py install` from the test suite, since these are deprecated. If preferred I can add something calling `python -m build`.

- I did not migrate the CFFI example to `pyproject.toml` because it uses a special `cffi_modules` keyword which is implemented by the `cffi` dependency itself and is not native to setuptools, with no `pyproject.toml`-equivalent. It is possible to migrate everything else to `pyproject.toml` and just leave `setup(cffi_modules=["cffi_module.py:ffi"])` behind, but I was not sure the maintainers would like the split. Please let me know if that should be changed.

- I assumed that the configuration in the `hello-world-script` example that uses a `RustExtension` with `Binding.Exec` is equivalent to a `RustBin` and turned the following
  ```python
  RustExtension(
      {"hello-world-script": "hello_world.hello-world-script"},
      binding=Binding.Exec,
      script=True,
      args=["--profile", "release-lto"],
  )
  ```
  into the TOML-equivalent:
  ```toml
  [[tool.setuptools-rust.bins]]
  target = "hello-world-script"
  args = ["--profile", "release-lto"]
  ```

---

#### Request to change the examples to use the same directory:

https://github.com/PyO3/setuptools-rust/pull/348#discussion_r1284176426:
> Personally I would prefer if we left the rust directory as src, because that's what the rest of the examples do. I think there's no good reason to diverge here, but I'm also happy to be persuaded that rust is better. If rust is agreed to be better, we should update the rest of the examples to use it too.

...
https://github.com/PyO3/setuptools-rust/pull/348#discussion_r1285257071
> Fair enough, let's keep rust directory here 👍

Not sure if there is an agreement about which directory is better, but it seems that the maintainer understand the controversy with the `src` directory and appreciated the effort to side-step this controversy[^1].

I intentionally left the commits implementing this change at the end of the PR so they can be reverted (or `git rest --hard`), if this change is not desirable.

[^1]: A "Python-first" developer would be confused with Rust files in `src` since the directory is historically used in Python packaging for Python files. A "Rust-first" developer would be confused with Python files in the `src` directory since Cargo also uses it as default. So the approach explored here follows the "explicit is better than implicit" mantra.

---
Please let me know if that is not what you had in mind and we can iterate over the implementation.

P.S.: I will work on changing the docs in a separated PR.


davidhewitt:
Thanks! I am happy with all the changes proposed here. You will probably need to change references in `.github/workflows/ci.yml` to use `python -m build` invocations instead of `setup.py`, as you noted elsewhere.

.github/workflows/ci.yml
```diff
@@ -269,11 +273,13 @@ jobs:
...
           PYO3_CROSS_LIB_DIR: /opt/python/cp38-cp38/lib
+          DIST_EXTRA_CONFIG: /tmp/build-opts.cfg
         run: |
           cd examples/namespace_package
           docker build -t cross-pyo3:aarch64-unknown-linux-gnu .
-          python -m pip install wheel
-          python setup.py bdist_wheel --plat-name manylinux2014_aarch64
+          python -m pip install build wheel
+          echo -e "[bdist_wheel]\nplat_name=manylinux2014_aarch64" > $DIST_EXTRA_CONFIG
+          python -m build --no-isolation
```
davidhewitt, line 282: Maybe the changes for `test-cross` and `test-crossenv` can be reverted and left for now to ease your pain in CI? Given `rust_with_cffi/setup.py` still exists I'd be ok with leaving those as-is in this PR and dealing with them another time.

.github/workflows/ci.yml
```diff
@@ -313,12 +319,14 @@ jobs:
...
+          DIST_EXTRA_CONFIG: /tmp/build-opts.cfg
         run: |
           mkdir -p $PYO3_CROSS_LIB_DIR
           docker cp -L $(docker create --rm quay.io/pypa/manylinux2014_aarch64:latest):/opt/python/cp38-cp38 /opt/python
           cd examples/namespace_package
-          python -m pip install wheel
-          python setup.py bdist_wheel --plat-name manylinux2014_aarch64
+          python -m pip install build wheel
+          echo -e "[bdist_wheel]\nplat_name=manylinux2014_aarch64" > $DIST_EXTRA_CONFIG
+          python -m build --no-isolation
```
davidhewitt, line 329: Potentially you could also rework this test to use the `rust_with_cffi` example and keep using `setup.py` for now?

davidhewitt (commented):
CI looking painful 😬

.github/workflows/ci.yml
abravalheri: Hi @davidhewitt, thank you very much for the review.

I can try to do that. For `test-crossenv` it should be just a matter of reverting 1d3fe5f32f95ac766bf4a53b67e85ef5e87674f8. Although, I am not sure how the problem with the failures in the CI are related to `setup.py` vs. `DIST_EXTRA_CONFIG` as a form of passing build options:

- In https://github.com/PyO3/setuptools-rust/actions/runs/5834731386/job/15824909577#step:3:232 we can see that the error in CI for test-crossenv seems unrelated (somehow it is missing `decimal`, which is supposedly happening before `build` creates an isolated environment.)

- In https://github.com/PyO3/setuptools-rust/actions/runs/5834731386/job/15824908490#step:7:488), we can see that Rust seems to be having problems to find Python information and failing with:
  ```python
  error: no Python 3.x interpreter found
  ```

Maybe that is related to the use of a virtual environment in https://github.com/PyO3/setuptools-rust/actions/runs/5834731386/job/15824908490#step:7:6?

I might try a couple of things before, but if they don't work I will proceed with the suggestion of sticking with `python setup.py` for the problematic tests.

abravalheri: There is something weird going on with `test-crossenv` when I try to use `build`:

```bash
+ pip install cffi
Requirement already satisfied: cffi in ./venv/build/lib/python3.9/site-packages (1.15.1)
Collecting pycparser
  Using cached pycparser-2.21-py2.py3-none-any.whl (118 kB)
Installing collected packages: pycparser
Successfully installed pycparser-2.21

Notice:  A new release of pip is available: 23.0.1 -> 23.2.1
Notice:  To update, run: pip install --upgrade pip
+ export DIST_EXTRA_CONFIG=/tmp/build-opts.cfg
+ DIST_EXTRA_CONFIG=/tmp/build-opts.cfg
+ echo -e '[bdist_wheel]\npy_limited_api=cp37'
+ python -m build --no-isolation
* Getting build dependencies for sdist...

ERROR Missing dependencies:
	cffi
```

I do `pip install cffi` and it says `Requirement already satisfied`, then I try to run `python -m build --no-isolation` and it complains about `cffi` being a missing dependency.

.github/workflows/ci.yml
abravalheri: @davidhewitt Since `distutils` impose as restriction that users cannot set `plat_name` directly on `build`, is it OK for `setuptools-rust` to be deriving its own value of `plat_name` from `build` when comparing the host and target architecture? Or I am misinterpreting the code?

abravalheri: It seems that changing the following [fixes the problem with `test-zigbuild`](https://github.com/PyO3/setuptools-rust/actions/runs/5857228514/job/15878634313?pr=352):

```diff
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -163,9 +163,11 @@ def add_rust_extension(dist: Distribution) -> None:
-                options = self.distribution.get_cmdline_options().get("bdist_wheel", {})
-                plat_name = options.get("plat-name") or self.plat_name
+                bdist_wheel = self.distribution.get_command_obj("bdist_wheel")
+                plat_name = bdist_wheel.plat_name or self.plat_name
```

(probably related to the fact that when the test calls `python -m build`, the `plat_name` is given via a `DIST_EXTRA_CONFIG` file instead of `sys.argv`, but I might be wrong).

.github/workflows/ci.yml
abravalheri: `crossenv` seems to have a problem with `build`: https://github.com/benfogle/crossenv/issues/108

davidhewitt (commented):
The `cibuildwheel` failure looks like it's because it's now building with pyproject.toml / PEP517 build isolation, which is causing the released `setuptools-rust` to be installed rather than the one from this branch.

Probably the most practical solution is to overwrite the PEP517 build requirements as a setup step in the cibuildwheel job, I think the below `sed` might work:

```
sed -i 's/requires = ["setuptools", "wheel", "setuptools-rust"]/requires = ["setuptools", "wheel", "${{ github.workspace }}"]/g" examples/namespace_package/pyproject.toml
```

----

I think the `cross` failure is caused because `python -m build` is building an sdist first, and the wheel from that, but the `Cross.toml` file is not included in the sdist. I guess solutions are either to include that file in the sdist, or to just do `python -m build --wheel` and skip the sdist.

abravalheri:
> The `cibuildwheel` failure looks like it's because it's now building with pyproject.toml / PEP517 build isolation, which is causing the released `setuptools-rust` to be installed rather than the one from this branch.

Thank you very much @davidhewitt, I think I managed to get `cibuildwheel` to work with build isolation (very tricky...[^1]), but now auditwheel is failing due to https://github.com/pypa/auditwheel/issues/436.

[^1]: When I use build isolation I have to upgrade setuptools first and only then install the working version of setuptools-rust.

> I think the cross failure is caused because python -m build is building an sdist first, and the wheel from that, but the Cross.toml file is not included in the sdist. I guess solutions are either to include that file in the sdist, or to just do python -m build --wheel and skip the sdist.

I went with adding it to `MANIFEST.in`.

setuptools_rust/setuptools_ext.py
```diff
@@ -163,13 +163,18 @@ def run(self) -> None:
...
                 build_rust.run()
 
             build_ext_base_class.run(self)
 
+        def _get_wheel_plat_name(self) -> Optional[str]:
+            try:
+                cmd = self.distribution.get_command_obj("bdist_wheel")
+                return cast(Optional[str], cmd.plat_name)
+            except Exception:  # unlikely scenario: `wheel` not installed
+                return None
```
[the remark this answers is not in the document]
abravalheri: For this specific scenario skipping `ensure_finalized` should be fine because `build_rust` already implements the same fallback for `plat_name` than `bdist_wheel` does (i.e. copying from `build`), so we are just interested in the value given via `sys.argv`/config file, which is available right after setuptools instantiates the command object inside `get_command_obj`.

But I agree that it would be nicer to share the logic between these 2 objects, so I will go ahead and implement it once I know what to do with the problem in auditwheel.

davidhewitt:
I think we might be about to get a green ci... 👀 

(This has been a truly heroic effort, thank you for all your help!)

davidhewitt (commented):
This is really great, thank you so much for this.

I think there's just the CHANGELOG entry to add for fixed plat name when it comes from `DIST_EXTRA_CONFIG` and also the refactoring of that code to be shared with `_py_limited_api`?

Once that's done, let's ship this in 1.7.0 ⛵

abravalheri:
Hi @davidhewitt I implemented the changes, but there seems to be something off with crates.io today [breaking the CI](https://github.com/PyO3/setuptools-rust/actions/runs/5900410048/job/16004621309?pr=352#step:8:105):

```
  Caused by:
    failed to query replaced source registry `crates-io`

  Caused by:
    download of 3/s/syn failed

  Caused by:
    failed to download from `[https://index.crates.io/3/s/syn`](https://index.crates.io/3/s/syn%60)
```

Other than that, my (hopefully) final remarks on this PR are:

- I started trying to only tackle the examples structure/config, but unfortunately it evolved to be much more than that and ended up being a PR to make the CI `build-isolation`-ready. Sorry for mixing the concerns :P 
- I left on the CI config some statements that clearly target debugging (e.g. increasing verbosity, `pip list` or `unzip -l dist/*.whl`). I found that these statements helped me to identify the root cause of the errors, and that in the future they might come handy again.
  Because of that I decided to leave them, for now, in the PR. But I am happy to remove them, please just let me know.

davidhewitt (approved):
This is brilliant as-is. Thank you very much for all the effort invested here; the CI and examples are in much better shape for all of this. I see you've also opened an issue with `cibuildwheel` to suggest further improvements upstream ❤️ 

I'll merge now and rebase the 1.7.0 release with a view to putting live early next week :)

PR #356, opened 2023-08-20
Update README/docs to use `pyproject.toml` + separate `setup.py` docs

abravalheri:
This PR attempts to target a few review comment in #348, specifically:

> We should also update documentation to cover (recommend?) this layout. Again, I'd be tempted to move the documentation of using setup.py to a secondary page in the docs and make this pyproject.toml form the preferred option.

Remarks:

- For the docs in the `README.md` I cannot use MyST-specific blocks (like `literalinclude`) or sphinx directives (like `{pep}`), because GitHub does not support it. So I just added a simplified version of the `hello-world` example, and linked where the users could obtain the `.rs` and `.py` files.
- I modified the `hello-world-setuppy` file to make it simpler and suitable for the separated docs for `setup.py`. The other examples already cover binaries, so I though that a shared library as Python module would be more relevant.
- I added a new `docs` session to `nox` to facilitate checking how the docs would be rendered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,83 +10,156 @@
 Compile and distribute Python extensions written in Rust as easily as if
 they were written in C.
 
-## Setup
+## Quickstart
 
-For a complete example, see
-[html-py-ever](https://github.com/PyO3/setuptools-rust/tree/main/examples/html-py-ever).
+The following is a very basic tutorial that shows how to use `setuptools-rust` in `pyproject.toml`.
+It assumes that you already have a bunch of Python and Rust files that you want
+distribute. You can see examples for these files in the 
+[`examples/hello-world`](https://github.com/PyO3/setuptools-rust/tree/main/examples/hello-world)
+directory in the [github repository](https://github.com/PyO3/setuptools-rust).
+[PyO3 docs](https://pyo3.rs) has detailed information on how to write Python
+modules in Rust.
 
-First, you need to create a bunch of files:
-
-### setup.py
-
-```python
-from setuptools import setup
-from setuptools_rust import Binding, RustExtension
-
-setup(
-    name="hello-rust",
-    version="1.0",
-    rust_extensions=[RustExtension("hello_rust.hello_rust", binding=Binding.PyO3)],
-    packages=["hello_rust"],
-    # rust extensions are not zip safe, just like C-extensions.
-    zip_safe=False,
-)
+```
+hello-world
+├── python
+│   └── hello_world
+│       └── __init__.py
+└── rust
+    └── lib.rs
 ```
 
-For a complete reference of the options supported by the `RustExtension` class, see the
-[API reference](https://setuptools-rust.readthedocs.io/en/latest/reference.html).
-
-### pyproject.toml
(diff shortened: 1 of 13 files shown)


docs/setuppy_tutorial.md
```diff
@@ -0,0 +1,130 @@
...
+your `setup.py`.
+
+
+## Basic implementation files
+
+Let's start by assuming that you already have a bunch of Python and Rust files[^1]
+that you would like to package for distribution in PyPI inside of a project directory
+named `hello-world-setuppy`[^2][^3]:
+
+[^1]: To now more about how to write Rust to be integrated into Python packages,
```
davidhewitt, line 18:
```suggestion
[^1]: To know more about how to write Rust to be integrated into Python packages,
```

davidhewitt (approved):
Thank you so much again, this is brilliant!

I was going to suggest we pause on merging this one until setuptools' `pyproject.toml` support is out of beta, and then I saw you already did that last week ❤️ 

Just a few tiny suggestions to the prose...

README.md
```diff
@@ -10,83 +10,156 @@
...
+binding = "PyO3"         # Default value, can be omitted
+py-limited-api = "auto"  # Default value, can be omitted
 ```
 
-### MANIFEST.in
-
-This file is required for building source distributions
+In this file we use an array of tables (TOML jargon equivalent to Python's list
+of dicts) for ``[[tool.setuptools-rust.ext-modules]]``, to specify different
+extension modules that will map directly to `Cargo.toml`'s `[lib]` table:
```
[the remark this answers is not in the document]
abravalheri: Thank you very much @davidhewitt for the review.

I tried to address this comment in a4e5c72. I hope that is inline with what you had in mind.

README.md
```diff
@@ -10,83 +10,159 @@
...
+## Quickstart
 
-For a complete example, see
-[html-py-ever](https://github.com/PyO3/setuptools-rust/tree/main/examples/html-py-ever).
+The following is a very basic tutorial that shows how to use `setuptools-rust` in `pyproject.toml`.
+It assumes that you already have a bunch of Python and Rust files that you want
+to distribute. You can see examples for these files in the 
+[`examples/hello-world`](https://github.com/PyO3/setuptools-rust/tree/main/examples/hello-world)
+directory in the [github repository](https://github.com/PyO3/setuptools-rust).
+The [PyO3 docs](https://pyo3.rs) have detailed information on how to write Python
```
abravalheri, line 20: Should we also mention `rust-cpython` here? Although I don't think there is any example using `rust-cpython` (or is there?)

README.md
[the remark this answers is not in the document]
davidhewitt: Perfect 👍

README.md
davidhewitt: I think it's fine as-is; `rust-cpython` is relatively unmaintained recently and as you say we also lack any example of it here. I think if someone wanted to contribute an example and extend documentation to discuss `rust-cpython` I would accept it, just no need to block this PR on it.

PR #357, opened 2023-08-23
raise minimum python version to 3.8 and drop typing_extensions

nekopsykose:
[typing.Literal](https://docs.python.org/3/library/typing.html#typing.Literal) was in python 3.8, so there isn't a need for this dependency on anything newer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
       # If one platform fails, allow the rest to keep testing if `CI-no-fail-fast` label is present
       fail-fast: ${{ !contains(github.event.pull_request.labels.*.name, 'CI-no-fail-fast') }}
       matrix:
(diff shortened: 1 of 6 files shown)


nekopsykose:
ah, mypy requires the specific Literal for itself..

davidhewitt:
Given that setuptools now requires 3.8, I'd also be ok with doing the same here to simplify things.

nekopsykose:
hm, that also works. sure

davidhewitt:
Thanks, looks like two final things needed here please:
- Remove 3.7 jobs from `.github/workflows/ci.yml`.
- Add a `Packaging` entry to the CHANGELOG (e.g. see release notes for 1.2 where we dropped 3.6 support).

nekopsykose:
good catch, thanks :)

davidhewitt (approved):
Perfect, thanks 👍

PR #364, opened 2023-10-03
Allow profile to be set by SETUPTOOLS_RUST_CARGO_PROFILE env variable

jefferyto:
This allows the profile to be set dynamically, without having to edit pyproject.toml/setup.py.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Drop support for Python 3.7. [#357](https://github.com/PyO3/setuptools-rust/pull/357)
 - Remove direct imports from `pkg_resources`. [#359](https://github.com/PyO3/setuptools-rust/pull/359)
 
+### Added
+- Add support for setting a custom cargo profile with the `SETUPTOOLS_RUST_CARGO_PROFILE` environment variable. [#364](https://github.com/PyO3/setuptools-rust/pull/364)
+
 ## 1.7.0 (2023-08-22)
 ### Packaging
 - Remove direct imports from `distutils`. [#336](https://github.com/PyO3/setuptools-rust/pull/336)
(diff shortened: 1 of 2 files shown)


davidhewitt:
I understand the desire for this (and have wanted it myself) but would like this to be called something other than `CARGO_PROFILE` as `cargo` itself does not recognise that environment variable. Perhaps a PR to `cargo` to implement this upstream would solve your use case and also help more users?

jefferyto:
> Perhaps a PR to `cargo` to implement this upstream would solve your use case and also help more users?

This would be better but I don't have time to wade into cargo internals right now :sweat_smile: I couldn't find any mention of this functionality in their issues/PRs, that is slightly surprising.

Also it turns out `CARGO_PROFILE` is already used to enable an [internal profiler](https://doc.crates.io/contrib/tests/profiling.html#internal-profiler), so that name is probably a no-go already.

Would `SETUPTOOLS_RUST_PROFILE` or `SETUPTOOLS_RUST_CARGO_PROFILE` be acceptable?

jefferyto:
I have changed the variable name to `SETUPTOOLS_RUST_CARGO_PROFILE`.

FYI I'm the Python maintainer at the OpenWrt packages feed; I'm working on improving the building of Rust extensions for the Python packages there. It would be much easier to patch setuptools-rust to accept this environment variable than to patch `args=["--profile=..."]` (and reading from an environment variable) into every Python package that needs it. (The profile selected would be determined by the OpenWrt build system.)

I'm not a Rust person so getting this isn't cargo isn't my first choice/priority right now.

davidhewitt:
Sorry for the delayed reply. Understood, in which case to help you move forward let's accept the environment variable here for now.

Can you please add a CHANGELOG entry, and then I'll get this merged and released asap.

jefferyto:
Thanks :pray: I've updated the CHANGELOG.

jefferyto:
Reformatted with black.

davidhewitt (approved):
Thanks! Will ship now 👍

PR #365, opened 2023-10-04
CI maintenance.

abravalheri:
This is a follow up on the previous CI changes.

- Avoid confusing `PIP_NO_BUILD_ISOLATION`.
  The latest version of `cibuildwheel` allows passing arguments directly to the build frontend CLI.
- Update `auditwheel` (via `pip`) to remove incompatibility with `wheel`'s latest version (due to de-vendoring of `auditwheel` by the container's OS package manager).
- Cap on `cffi<1.16` for mingw.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -312,7 +312,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           targets: aarch64-apple-darwin
-      - uses: pypa/cibuildwheel@v2.3.1
+      - uses: pypa/cibuildwheel@v2.16.2
         env:
           CIBW_BUILD: cp39-*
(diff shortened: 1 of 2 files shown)


abravalheri:
Hi @davidhewitt, it seems that cffi 1.16 started breaking for mingw...
I noticed that in https://github.com/PyO3/setuptools-rust/actions/runs/6404823991/job/17386133529, a completely unrelated change in the `cibuildwheel` workflow.

davidhewitt (approved):
👍 thanks for this!

Capping cffi seems fine for now, can dig another time when it comes to it, ultimately it shouldn't matter for us in CI.

abravalheri:
I opened https://github.com/python-cffi/cffi/issues/21, but maybe the problem is in setuptools itself.
Any suggestions are welcome.

PR #382, opened 2023-11-17
fix readthedocs configuration

davidhewitt:
The 1.8.x releases are missing from readthedocs.io, I think it's due to an invalid configuration file.


davidhewitt:
Builds are now running but failing, probably the docs dependencies are wildly outdated, I'll merge this and add dependabot for the docs in a separate PR.

PR #390, opened 2023-11-18
Fix CI with new cargo workspace

davidhewitt:
I added a cargo workspace for the examples in #388, which is a little gross but IMO necessary to ensure that dependabot configuration reliably picks up all the examples.

This should fix CI following that tweak.


davidhewitt:
Ok, the workspace seems to be more trouble than it's worth, so I've reverted to enumerating examples in dependabot config and removed the workspace again.

PR #400, opened 2024-01-20
Update for ABI builds in pyproject.toml

### LINKED ISSUE #399: Trouble using pyproject.toml for limited api
```
python setup.py bdist_wheel --py-limited-api=cp38
```

Seems to work fine, but I try to use `python -m build` where the toml is configured with the extension, then it doesn't build with the limited api.

```
[[tool.setuptools-rust.ext-modules]]
target = "mytarget"
py-limited-api = true
```

I put a breakpoint down in the code, and this is the point where I can tell there is a difference and the cmd_obj ends up with `cmd.py_limited_api = False` using `build` only, but in the working case it is `cp38`

```python

def _get_bdist_wheel_cmd(  # type: ignore[no-any-unimported]
    dist: Distribution, create: bool = True
) -> Optional[CommandBdistWheel]:
    breakpoint()
    try:
        cmd_obj = dist.get_command_obj("bdist_wheel", create=create)
        cmd_obj.ensure_finalized()  # type: ignore[union-attr]
        return cast(CommandBdistWheel, cmd_obj)  # type: ignore[no-any-unimported]
    except Exception:
        return None

```

krpatter-intc:
Closes #399

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,11 @@ target = "hello_world._lib"  # The last part of the name (e.g. "_lib") has to ma
 path = "Cargo.toml"      # Default value, can be omitted
 binding = "PyO3"         # Default value, can be omitted
 py-limited-api = "auto"  # Default value, can be omitted
+
+# Optional, only needed if building with limited-api
+# Be sure to update with the desired version
+[tool.distutils.bdist_wheel]  # Optional
+py_limited_api = "cp38"       # Optional
 ```
 
 Each extension module should map directly into the corresponding `[lib]` table on the


davidhewitt:
This looks useful! Have you got a link to documentation for distutils about this `tool.distutils` toml section? I'd just like to understand any stability or additional features before we recommend this in docs.

krpatter-intc:
Well, the wheel docs are extremely sparse:

https://wheel.readthedocs.io/en/stable/

I found the setup.cfg documented here but they don't mention anything about pyproject.toml
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#wheels

But...I can't find anything officially about setting the universal wheel setting either...seems yet another spot in the ecosystem where pyproject.toml is _barely_ supported.

davidhewitt:
I see, thanks. I'm a bit wary of recommending the `tool.distutils` section of the `pyproject.toml` in our docs if we can't find any upstream in it.

I might prefer instead to mention `DIST_EXTRA_CONFIG` like this for now: https://github.com/PyO3/setuptools-rust/pull/410/commits/e3db718fa9906d53fa4fd27f29ba806c0680f893

davidhewitt:
Based on having merged the other PR, I'll close this - thanks again for prompting the discussion and sharing this solution 🙏

PR #410, opened 2024-02-01
deprecate `RustExtension.py_limited_api`

davidhewitt:
I'd like to simplify the `RustExtension` options a little further and remove the `py_limited_api` setting, as I think users should always use `"auto"` and use `[bdist_wheel]` configuration to set up an abi3 build.

If the PyO3 module should be built in abi3 mode for local (i.e. non-wheel) builds too, then that configuration is probably better placed in the `Cargo.toml` than in the `setuptools-rust` layer.

cc @alex as you implemented this setting originally, but I think `cryptography` has now also switched to `auto` mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+### Changed
+- Deprecate `py_limited_api` option to `RustExtension` in favour of always using `"auto"` to configure this from `bdist_wheel`. [#410](https://github.com/PyO3/setuptools-rust/pull/410)
+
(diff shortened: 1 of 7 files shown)


alex:
Yes, we now use auto + `features = ["abi3"]` in Cargo.toml + `--config-settings=--build-option=--py-limited-api=${{ matrix.PYTHON.ABI_VERSION }}` when building release wheels.

Which I think means this is fine for us.

davidhewitt:
Great! I should probably also add the `--config-settings` option to the docs too 👍

PR #429, opened 2024-05-23
Use base executable when it exists, to avoid recompiling when switching venv

matsjoyce-refeyn:
When using a requirements file compilation tool like pip-tools, a lot of temporary virtual environments are created, into which packages are installed. When a package uses setuptools-rust, the value of `sys.executable` keeps changing, which cause cargo to recompile quite a few crates (pyo3 and anything that depends on it). This slows things down quite a lot.

This PR uses the [semi-documented](https://docs.python.org/3/c-api/init_config.html#c.PyConfig.base_executable) `sys._base_executable` value, which points to the real Python interpreter when inside a venv. If it doesn't exist, or the provided path is not valid, we fall back to the current behaviour.

I tested this by running a requirements compilation, and checking for rustc invocations: There are a lot before this change, and none after (provided the rust package in question was already compiled).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+### Changed
+ - Use the base interpreter path when running inside a virtual environment to avoid recompilation when switching between virtual environments. [#429](https://github.com/PyO3/setuptools-rust/pull/429)
+
 ## 1.9.0 (2024-02-24)
 ### Changed
 - Deprecate `py_limited_api` option to `RustExtension` in favour of always using `"auto"` to configure this from `bdist_wheel`. [#410](https://github.com/PyO3/setuptools-rust/pull/410)
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -610,9 +610,13 @@ def _replace_vendor_with_unknown(target: str) -> Optional[str]:
 def _prepare_build_environment() -> Dict[str, str]:
     """Prepares environment variables to use when executing cargo build."""
 
+    executable = getattr(sys, "_base_executable", sys.executable)
+    if not os.path.exists(executable):
+        executable = sys.executable
+
     # Make sure that if pythonXX-sys is used, it builds against the current
     # executing python interpreter.
-    bindir = os.path.dirname(sys.executable)
+    bindir = os.path.dirname(executable)
 
     env = os.environ.copy()
     env.update(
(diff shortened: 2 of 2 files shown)


davidhewitt:
Thanks for the PR! This diff seems very reasonable to me, and definitely a desirable goal to reduce recompilation.

A couple of design questions:
- Are there any downsides we can think of by doing this? For example, PyO3 doesn't yet try to bake information about virtualenvs into binaries. Should it? My instinct is _probably not_ (and it should use runtime detection), but we have known gaps in virtualenv support in PyO3 and I wonder if this change interacts with that.
- Rather than using `_base_executable`, should we use the relative difference between `sys.prefix` and `sys.executable`, and apply that to `sys.base_prefix`? Again, probably not a good idea, though I wanted to at least put the idea out for discussion.

So overall this seems reasonable to me. We should add a note in the CHANGELOG for this please.

matsjoyce-refeyn:
> * Are there any downsides we can think of by doing this? For example, PyO3 doesn't yet try to bake information about virtualenvs into binaries. Should it? My instinct is _probably not_ (and it should use runtime detection), but we have known gaps in virtualenv support in PyO3 and I wonder if this change interacts with that.

My assumption is that since PyO3 (in this case) cares about Python from a compilation POV, virtual environments don't affect it as all the interesting header-related things are not touched by the venv.

From what I read online, the issues other users were having with PyO3 was to do with activating a venv when running Python embedded in Rust, so this would have no effect (good or bad) on that. But my search was not very thougher, are there any other known issues?

> * Rather than using `_base_executable`, should we use the relative difference between `sys.prefix` and `sys.executable`, and apply that to `sys.base_prefix`? Again, probably not a good idea, though I wanted to at least put the idea out for discussion.

We could, my worry would be that there are a lot of special cases, since I don't think there's a requirement that the name of the executable matches (that seems to have been the problem with https://github.com/python/cpython/issues/99204). By using `sys._base_executable`, we offload that problem onto the stdlib authors, who have a better chance of thinking though all the possibilities than I have 😄 

> So overall this seems reasonable to me. We should add a note in the CHANGELOG for this please.

Will do!

davidhewitt (approved):
Thanks 👍

messense:
This does not seem like a reliable solution, I've tried in [maturin](https://github.com/PyO3/maturin/pull/2094), when using with poetry to install an editable pyo3 based dep,

```bash
$ /tmp/tmp3uzvm48r/.venv/bin/python
Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0] on linux
>>> import sys
>>> sys.executable
'/tmp/tmp3uzvm48r/.venv/bin/python'
>>> sys._base_executable
'/tmp/tmp3uzvm48r/.venv/bin/python'
```

the isolated venv poetry creates has `sys.executable == sys._base_executable`.

PR #433, opened 2024-06-03
fix mypy error on main

davidhewitt:


setuptools_rust/build.py
```diff
@@ -786,7 +786,7 @@ def _replace_cross_target_dir(path: str, ext: RustExtension, *, quiet: bool) ->
 
 
 def _get_bdist_wheel_cmd(  # type: ignore[no-any-unimported]
-    dist: Distribution, create: bool = True
+    dist: Distribution, create: Literal[True, False] = True
```
adamreichold, line 789: Is this implied by a mypy change? `bool` does look more reasonable than `Literal[True, False]` even if essentially equivalent.

davidhewitt: Based on the error it seems like `get_command_obj` has two overloads, which take either `Literal[0, False]` or `Literal[1, True]`. I don't know whether it's a mypy change or new hints shipped in setuptools which is causing `bool` to not fit well into these overloads.

I agree that `bool` was better but overall as this was an internal function I kinda just wanted to suck it up and silence the warning.

adamreichold (approved):

PR #435, opened 2024-06-05
Canonicalize base executable path when finding interpreter

matsjoyce-refeyn:
Port across the change from https://github.com/PyO3/maturin/pull/2100 (based on https://github.com/PyO3/setuptools-rust/pull/429#issuecomment-2148702096) which allows this improvement to work with Poetry (thanks @messense!).


messense (approved):

davidhewitt (approved):
Thanks! 👍

PR #436, opened 2024-06-05
iOS support

FeodorFitsner:
A tiny change that makes `setuptools-rust` work in `crossenv` environment, originally made by @freakboy3742
It allows to successfully build `bcrypt` and `cryptography` for iOS.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -190,8 +190,8 @@ class build_rust(RustCommand):
                     *ext.rustc_flags,
                 ]
 
-            # OSX requires special linker arguments
-            if rustc_cfgs.get("target_os") == "macos":
+            # Apple platforms require special linker arguments
+            if rustc_cfgs.get("target_os") in {"macos", "ios", "tvos", "watchos"}:
                 ext_basename = os.path.basename(self.get_dylib_ext_path(ext, ext.name))
                 rustc_args.extend(
                     [


freakboy3742:
FYI - I was deliberately *not* submitting this PR upstream until such time a I had an opportunity to fully test it with the final state of iOS support in CPython 3.13, and add tests to verify the change. 

On a purely functional level, while it successfully *builds* cryptography, in my testing the final binary doesn't fully work, as there's a runtime load that fails, and I wasn't 100% sure if that would require additional changes to this package.

I don't object to this being merged as is if the maintainers are comfortable with it; at least to my mind, there's more work required before it is ready.

FeodorFitsner:
Any chance to get this change into the next release of setuptools-rust?

davidhewitt:
I will be working on the next release of setuptools-rust imminently now that we have 3.13 support in PyO3. So yes, am happy to move forward with this.

Do you know if there's any reasonable way we can test iOS in CI? If not, no problem, let's add a CHANGELOG entry and then merge this as-is 👍

freakboy3742:
It's definitely possible to build/test packages on iOS; for example [here's a CI test run for Toga (a GUI toolkit) running on iOS](https://github.com/beeware/toga/actions/runs/9687179957/job/26731313255). You need to use a macOS runner; the compilation process is no slower than it would be for iOS, and the SDKs are all installed and available. Running the code means starting a simulator; that takes 2-3 minutes by itself, so it's not *fast*, but it's very reliable.

My Q3 TODO list is largely focussed on binary packaging for iOS; I should be in a position to look at this in a couple of weeks. I don't know if that means you want to hold off merging until CI is in place.

PR #437, opened 2024-06-10
Delay imports of non-stdlib dependencies until time of use

eli-schwartz:
This is a bit of a hack, but we do this to ensure that when setuptools loads entrypoint based hooks it cannot (will not?) crash.

The issue is that setuptools plugins are autoloaded, whether any given project uses them at all or not. So if setuptools-rust is installed, setuptools always tries to use it, and crashes if setuptools-rust is broken.
    
Of course, setuptools-rust can't be broken, because it's a wonderful project.

BUT.

As it happens, third-party vendors providing setuptools-rust can get into a situation where multiple packages need to be installed, including setuptools-rust, and also build yet other packages from source. In the middle of this, setuptools-rust itself could be installed but in "half-configured" state, i.e. its dependencies were queued for afterwards due to complicated dependency graph magic.

In such a scenario, it should be nominally all right to have an inert package installed, since if nothing actually uses setuptools-rust it doesn't need to *work* yet. And in fact, it is all right, as long as setuptools can import the autoloaded plugin hooks (and do nothing with them).

Bug: https://bugs.gentoo.org/933553

### DIFF
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -26,7 +26,10 @@ except ImportError:
 if sys.version_info[:2] >= (3, 11):
     from tomllib import load as toml_load
 else:
-    from tomli import load as toml_load
+    try:
+        from tomli import load as toml_load
+    except ImportError:
+        from setuptools.extern.tomli import load as toml_load
 
 
 logger = logging.getLogger(__name__)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 ### Changed
  - Use the base interpreter path when running inside a virtual environment to avoid recompilation when switching between virtual environments. [#429](https://github.com/PyO3/setuptools-rust/pull/429)
+ - Delay import of dependencies until use to avoid import errors during a partially complete install when multiple packages are installing at once. [#437](https://github.com/PyO3/setuptools-rust/pull/437)
 
 ## 1.9.0 (2024-02-24)
 ### Changed
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ classifiers = [
 dependencies = [
     "setuptools>=62.4",
     "semantic_version>=2.8.2,<3",
-    'tomli>=1.2.1; python_version<"3.11"'
 ]
 
 [project.entry-points."distutils.commands"]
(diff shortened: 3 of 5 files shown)


setuptools_rust/setuptools_ext.py
```diff
@@ -26,7 +26,10 @@
 if sys.version_info[:2] >= (3, 11):
     from tomllib import load as toml_load
 else:
-    from tomli import load as toml_load
+    try:
+        from tomli import load as toml_load
+    except ImportError:
+        from setuptools.extern.tomli import load as toml_load
```
davidhewitt, line 32: Interesting; I wasn't aware of this namespace and have just asked upstream at https://github.com/pypa/setuptools/pull/4351#issuecomment-2159316805

davidhewitt (commented):
Thanks for the PR! I think the motivation is very reasonable, so let's find a way to support this edge case.

To keep the code simpler, are you perhaps willing to adjust this PR to vendor (with appropriate attribution) the parts of `semantic_version` we actually use? I think we use barely any of it, and maybe a nicer end state is for `setuptools_rust` to have no external dependencies except `setuptools`?

eli-schwartz:
> To keep the code simpler, are you perhaps willing to adjust this PR to vendor (with appropriate attribution) the parts of `semantic_version` we actually use? I think we use barely any of it, and maybe a nicer end state is for `setuptools_rust` to have no external dependencies except `setuptools`?

It's not a huge project. It's 1185 lines of code and looks like it could trivially reduce to 835, which seems hairy.

And I don't really think dependencies are *bad* per se, at least when you actually intend to use them.

OTOH I have thought for years that version parsing should be part of the stdlib, but the only move on that part was to remove distutils.version without stdlib-replacement as part of removing distutils. :shrug: Even just PEP 440 support would be nice considering it's such a building block for other python tools and has no real churn.

davidhewitt (approved):
Understood; let's proceed with this approach then!

Zero dependencies would have been nice for ease of testing, though I agree the `semantic_version` code doesn't reduce as nicely as I would have liked.

eli-schwartz:
Thanks! I've backported this PR now, so that users doing upgrades with complex dependency graphs on Gentoo can have the fix for 1.9.0 as well.

thesamesam:
Many thanks for this both!

davidhewitt:
Great! I will probably release 1.10.0 soon (maybe with testing on 3.13).

PR #448, opened 2024-07-01
release: 1.10.0

davidhewitt:


alex:
I'm concerned that #429 is going to break cryptography's build.

https://github.com/pyca/cryptography/blob/main/src/rust/cryptography-cffi/build.rs#L29-L37

This relies on running in the same venv where the build requirements were installed, because that Python script imports cffi.

Not sure if there's a way to write that script that's compatible with this change (which seems quite deserirable!)

davidhewitt:
Hmm, good point. I think in general the fact that #429 is unconditionally breaking build isolation seems problematic

cc @messense; might also be a problem for maturin?

Not sure what to do here, the obvious solution is to revert #429 but maybe there's a better solution as you say. Avoiding needless recompilation would be great. Probably we have to move towards `PYO3_ENVIRONMENT_SIGNATURE` like `maturin` does.

Either way, `setuptools-rust` ran fine on 3.13 without changes so holding up the release a little bit while we work this out isn't going to block 3.13 testing.

messense:
> cc @messense; might also be a problem for maturin?

yes, we've got a report in maturin related to this, I've changed it to require setting a env var to opt-in the behavior in https://github.com/PyO3/maturin/pull/2134

davidhewitt:
Ok, I think this should be better now with #458 merged.

alex:
Sounds great.

davidhewitt:
This is now live 👍

PR #460, opened 2024-08-04
release: 1.10.1

### LINKED ISSUE #459: `python -m pip install -e .` builds `release` instead of `debug`
I've started to get issues in CI, after debugging I can say that this commit
https://github.com/PyO3/setuptools-rust/commit/925b3ba4186ccd53ee8d123c07e03e60558f13b5
introduces changes that breaks previous behavior:
after running
```bash
> python -m pip install -e .
```
previously `target/debug` build was created and after that commit it started creating `target/release` instead, is it expected?

davidhewitt:
Pushed a commit to fix #459, and then the release commit.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,7 +1,7 @@
 [bumpversion]
 commit = True
 tag = False
-current_version = 1.10.0
+current_version = 1.10.1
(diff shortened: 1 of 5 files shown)


davidhewitt:
Release is live

PR #462, opened 2024-08-24
fix: pass RUSTFLAGS

n-bes:
Custom `RUSTFLAGS` passed only when internal `RUSTFLAGS` is set, fixed it. Possibly fix https://github.com/pypa/pip/issues/12935

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -231,7 +231,7 @@ class build_rust(RustCommand):
             ]
 
         # order is important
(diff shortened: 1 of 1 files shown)


n-bes:
Oh, it is already copied...
https://github.com/PyO3/setuptools-rust/blob/main/setuptools_rust/build.py#L625

davidhewitt (commented):
Thanks for the PR. I'm not sure I understand the motivation for the patch, nor what the functional change is from the patch submitted. Could you please elaborate?

PR #463, opened 2024-08-28
Handle `wheel.bdist_wheel` deprecation warning

### LINKED ISSUE #461: `wheel.bdist_wheel` is depracated
The `wheel.bdist_wheel` has been deprecated in `wheel 0.44.0` and so it should be replaced by `setuptools.command.bdist_wheel` from `setuptools>=70.1`.

abravalheri:
... by trying to import first from setuptools.

I also changed some of the test/CI related files to remove the (no-longer-)dependency on `wheel` (which also implies `setuptools>=70.1` in some cases).

Finally it was necessary to add a temporary workaround for python-cffi/cffi#117 (i.e. "pin" `setuptools<74` for some scenarios). This workaround can be removed once python-cffi/cffi#117 is tackled.

Closes #461

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
           set -e
           cd examples/rust_with_cffi/
           python --version
-          pip install -U build cffi 'setuptools>=70.1,<74'  # TODO: Remove pin once python-cffi/cffi#117 is solved
+          pip install -U build cffi 'setuptools>=70.1'
           echo -e "[bdist_wheel]\npy_limited_api=cp39" > $DIST_EXTRA_CONFIG
           python -m build --no-isolation
           ls -la dist/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
           rustup target add aarch64-apple-darwin
           rustup target add x86_64-apple-darwin
           cd examples/namespace_package
-          pip install build wheel
+          pip install build
           python -m build --no-isolation
           ls -l dist/
           pip install --force-reinstall dist/namespace_package*_universal2.whl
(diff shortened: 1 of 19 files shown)


setuptools_rust/setuptools_ext.py
```diff
@@ -19,9 +19,12 @@
...
 
 try:
-    from wheel.bdist_wheel import bdist_wheel
+    from setuptools.command.bdist_wheel import bdist_wheel
 except ImportError:
-    bdist_wheel = None
+    try:  # old version of setuptools
+        from wheel.bdist_wheel import bdist_wheel
+    except ImportError:
+        bdist_wheel = None
```
abravalheri, line 27: This is a bit ugly (too nested), but similar to other code in the same file...
Alternatively we could also shove this conditional import into a separated file to hide the mess.

setuptools_rust/build.py
```diff
@@ -37,9 +37,9 @@
 
 
 try:
-    from wheel.bdist_wheel import bdist_wheel as CommandBdistWheel
-except ImportError:  # wheel installation might be deferred in PEP 517
-    from setuptools import Command as CommandBdistWheel
+    from setuptools.command.bdist_wheel import bdist_wheel as CommandBdistWheel
+except ImportError:  # old version of setuptools
+    from setuptools import Command as CommandBdistWheel  # type: ignore[assignment]
```
eli-schwartz, line 42: You should instead try all 3 imports...

eli-schwartz (changes requested):

setuptools_rust/build.py
abravalheri: No problems, I can do that. I skipped the `wheel` import in this particular case because it is used for type checking.

abravalheri: Implemented in [2e1d65b](https://github.com/PyO3/setuptools-rust/pull/463/commits/2e1d65baa206f1aaea6631a47b788631b6f5a1bc).

.github/workflows/ci.yml
```diff
@@ -163,7 +163,7 @@ jobs:
           set -e
           cd examples/rust_with_cffi/
           python --version
-          pip install -U build cffi wheel
+          pip install -U build cffi 'setuptools>=70.1,<74'  # TODO: Remove pin once python-cffi/cffi#117 is solved
```
davidhewitt, line 166: I think cffi patch is now released:

```suggestion
          pip install -U build cffi 'setuptools>=70.1'
```

davidhewitt (approved):
Thanks as always, looks good to me!

davidhewitt:
(Needs a CHANGELOG entry; I will push one and add later if you don't beat me to it 👍)

PR #502, opened 2024-12-05
add free-threading support

davidhewitt:
Let's add to CI and see what happens! cc @ngoldbaum 

(mingw builds are expected to be broken for unrelated reasons, fixed separately in #501)


davidhewitt:
Looks like this just straight up worked. I'll cut a release anyway to signal explicit support to give users confidence.

PR #503, opened 2024-12-06
ci: fix mingw with workaround

davidhewitt:
Cut down version of #501 which should enable CI to be green while waiting on the long-term fix in `nox`.


davidhewitt:
#501 was merged.

PR #504, opened 2025-01-01
Support adding custom env vars

konstin:
Currently, maturin and setuptools-rust are not self-contained: You need to install rust on the machine in order to build a package. We can improve this by downloading a matching rustup, using it to install cargo and rustc into a temporary directory and using this rust for the build. As long as the platform is supported by rustup, we get a self-contained build that doesn't mess with the host system. You can for example use `pip install ...` in a fresh `python3-slim` docker container and any rust dependency with a missing wheel will build and install without additional setup.

To inject this cache installation, we need to set `PATH`, `RUST_HOME` and `CARGO_HOME` whenever we call `cargo` or `rustc`.

With these changes, we can make bootstrapping maturin self-contained:

```python
if not os.environ.get("MATURIN_NO_INSTALL_RUST") and not shutil.which("cargo"):
    from puccinialin import setup_rust

    print("Rust not found, installing into a temporary directory")
    extra_env = setup_rust()
    env = {**os.environ, **extra_env}
else:
    env = None

setup(
    version=version,
    cmdclass={"bdist_wheel": bdist_wheel},
    rust_extensions=[RustBin("maturin", args=cargo_args, cargo_manifest_args=["--locked"], env=env)],
    zip_safe=False,
)
```

```python
"""Support installing rust before compiling (bootstrapping) maturin.

Installing a package that uses maturin as build backend on a platform without maturin
binaries, we install rust in a cache directory if the user doesn't have a rust
installation already. Since this bootstrapping requires more dependencies but is only
required if rust is missing, we check if cargo is present before requesting those
dependencies.

https://setuptools.pypa.io/en/stable/build_meta.html#dynamic-build-dependencies-and-other-build-meta-tweaks
"""

from __future__ import annotations

import os
import shutil
from typing import Any

# noinspection PyUnresolvedReferences
from setuptools.build_meta import *  # noqa:F403


def get_requires_for_build_wheel(_config_settings: dict[str, Any] = None) -> list[str]:
    if not os.environ.get("MATURIN_NO_INSTALL_RUST") and not shutil.which("cargo"):
        return ["puccinialin"]
    return []


def get_requires_for_build_sdist(_config_settings: dict[str, Any] = None) -> list[str]:
    if not os.environ.get("MATURIN_NO_INSTALL_RUST") and not shutil.which("cargo"):
        return ["puccinialin"]
    return []
```

### DIFF
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -15,6 +15,7 @@ from setuptools.command.install_scripts import install_scripts
 from setuptools.command.sdist import sdist
 from setuptools.dist import Distribution
 
+from ._utils import Env
 from .build import _get_bdist_wheel_cmd
 from .extension import Binding, RustBin, RustExtension, Strip
 
@@ -96,7 +97,15 @@ def add_rust_extension(dist: Distribution) -> None:
                 #
                 # https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options
                 cargo_manifest_args: Set[str] = set()
+                env: Optional[Env] = None
                 for ext in self.distribution.rust_extensions:
+                    if env is not None:
+                        if ext.env != env:
+                            raise ValueError(
+                                "For vendoring, all extensions must have the same environment variables"
+                            )
+                    else:
+                        env = ext.env
                     manifest_paths.append(ext.path)
                     if ext.cargo_manifest_args:
                         cargo_manifest_args.update(ext.cargo_manifest_args)
@@ -120,7 +129,7 @@ def add_rust_extension(dist: Distribution) -> None:
                     # set --manifest-path before vendor_path and after --sync to workaround that
                     # See https://docs.rs/clap/latest/clap/struct.Arg.html#method.multiple for detail
                     command.extend(["--manifest-path", manifest_paths[0], vendor_path])
-                    subprocess.run(command, check=True)
+                    subprocess.run(command, check=True, env=env.env if env else None)
 
                     cargo_config = _CARGO_VENDOR_CONFIG
 
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -98,14 +98,19 @@ def add_rust_extension(dist: Distribution) -> None:
                 # https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options
                 cargo_manifest_args: Set[str] = set()
                 env: Optional[Env] = None
+                env_source: Optional[str] = None
                 for ext in self.distribution.rust_extensions:
                     if env is not None:
                         if ext.env != env:
                             raise ValueError(
-                                "For vendoring, all extensions must have the same environment variables"
+                                f"For vendoring, all extensions must have the same environment variables, "
+                                f"but {env_source} and {ext.name} differ:\n"
+                                f"{env_source}: {env}\n"
+                                f"{ext.name}: {ext.env}"
                             )
                     else:
                         env = ext.env
+                        env_source = ext.name
                     manifest_paths.append(ext.path)
                     if ext.cargo_manifest_args:
                         cargo_manifest_args.update(ext.cargo_manifest_args)
--- a/setuptools_rust/setuptools_ext.py
+++ b/setuptools_rust/setuptools_ext.py
@@ -1,5 +1,4 @@
 import os
-import subprocess
 import sys
 import sysconfig
 import logging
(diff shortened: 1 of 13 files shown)


konstin:
The mingw CI failures are also on main

davidhewitt (commented):
Thanks! Implementation looks fine, though I slightly worry about the complexity.

I tried to look for prior art, as far as I can tell Cython / setup tools doesn't have a supported option and just leaves you to modify `os.environ` as part of `setup.py`. Is there a reason why that's not good enough for this use case here?

davidhewitt:
(Mingw failures I've just resolved in #501. PyPy 3.10 is just flat out broken as far as I can tell until the next PyPy version is released.)

konstin:
> I tried to look for prior art, as far as I can tell Cython / setup tools doesn't have a supported option and just leaves you to modify os.environ as part of setup.py. Is there a reason why that's not good enough for this use case here?

`os.environ` is global state and i'd prefer to isolate those values, as far as that's possible for environment variables. In my experience that helps with debugging.

> Mingw failures I've just resolved in https://github.com/PyO3/setuptools-rust/pull/501. PyPy 3.10 is just flat out broken as far as I can tell until the next PyPy version is released.

Thanks I'll rebase.

setuptools_rust/setuptools_ext.py
```diff
@@ -96,7 +97,15 @@ def make_distribution(self) -> None:
                 #
                 # https://doc.rust-lang.org/cargo/commands/cargo-build.html#manifest-options
                 cargo_manifest_args: Set[str] = set()
+                env: Optional[Env] = None
                 for ext in self.distribution.rust_extensions:
+                    if env is not None:
+                        if ext.env != env:
+                            raise ValueError(
+                                "For vendoring, all extensions must have the same environment variables"
```
davidhewitt, line 105: It might be nice to improve this message with something like "Extension('foo') has explicit env passed as an option, but other extensions had different environment (or no environment)"

Not sure that is exactly the right wording, but it might help users understand what they did wrong? 🤔

davidhewitt (commented):
> `os.environ` is global state and i'd prefer to isolate those values, as far as that's possible for environment variables. In my experience that helps with debugging.

I agree with this, though I am also willing to be a bit more lax for simplicity's sake in these "build system" components especially when it's sort of ok (IMO) to treat `setup.py` as an entrypoint setting up build configuration. 

I guess the main cost of the isolation (as we see in this PR) is that we have to be careful to pass the env dict to all downstream processes; forgetting to do so once is probably easy to miss in review and will create bugs. Can we use `ruff` to forbid use of subprocess outside our own private wrapper which requires an `Env`? (Maybe [this setting](https://docs.astral.sh/ruff/settings/#__tabbed_94_1)?)

Is there any place outside of subprocesses where these variables might matter? I think it seems clear enough from docs that we only intend for this to affect rustc / cargo. The only other subprocesses we call (as far as I see from a quick scan) are `strip` and `lipo` (on various platforms), but I think there is little need for those to be affected by this option.

I think as long as we have ways to avoid this getting out of sync later with some tests or lints, this gets 👍 from me.

konstin:
That's a though problem, i can't figure out a good solution: The subprocess module has a number of functions that create a subprocess, and we don't want to ban using them, we only want to ensure that `env=` is set explicitly so that you don't accidentally forget the environment (or set `env=None` if it's not a rustc/cargo call). I don't see how to do that with `lint.flake8-tidy-imports`, did you have something specific in mind?

cassiersg:
As an independent use-case for this feature: I just needed this feature today: I am building two `RustExtension`s in a single `setup(...)`, and the extensions need different `RUSTFLAGS` env vars. As far as I can tell, this is currently impossible. (I ended up monkey-patching `_prepare_build_environment` with some `inspect` magic.)

cassiersg:
@konstin The `subprocess` doc actually says:

> The recommended approach to invoking subprocesses is to use the [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) function for all use cases it can handle. For more advanced use cases, the underlying [Popen](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) interface can be used directly.

Therefore, it might not be an issue to ban the other functions (the ["older high-level API"](https://docs.python.org/3/library/subprocess.html#older-high-level-api). Then, it remains to provide a wrapper for `run` as proposed by @davidhewitt  (and maybe ban `Popen` - or we consider that it's use is unlikely / shows explicit intent).

davidhewitt:
Sorry for the long delay, I pushed a commit to at the lint rules as I wanted them. I'll wait for CI to run, and if it's green I'll merge and release. 

Thanks for implementing & for the patience! 🙏

davidhewitt:
`test-cross` is failing on `main` so will fix separately.

PR #517, opened 2025-03-14
fix `test-cross` build failure

davidhewitt:
I _think_ what might be going on is that `cargo metadata` is writing output to `stderr` which is getting caught in `check_output`; let's see.


davidhewitt:
Actually looks like the issue is just https://github.com/cross-rs/cross/issues/1645, looks like it will sort itself, will not try to change anything here.

PR #521, opened 2025-03-28
Fix finding cargo artifacts when `filenames` are empty

messense:
For https://github.com/PyO3/maturin/issues/2370, we used to have this in maturin before switching to `setuptools-rust` for bootstrap.


davidhewitt:
Thanks, I see no harm in merging and shipping this 👍

PR #527, opened 2025-04-04
use `SETUPTOOLS_RUST_CARGO_PROFILE=dev` for main matrix, document it

davidhewitt:
This will make CI run a little faster maybe, and also helps users discover the flag.


davidhewitt:
Only helps by maybe ~20 seconds per job, but that's still going to add up across the whole matrix (and we're short on CI runner time across the PyO3 org at the moment).

PR #534, opened 2025-05-18
Strip target suffix for cargo-zigbuild compatibility

bet4it:
When using `cargo-zigbuild`, target triples may include additional suffixes like glibc versions (e.g. `aarch64-unknown-linux-gnu.2.17`). This change strips any suffix after the dot to ensure compatibility with rustc, which only accepts standard target triples.


davidhewitt:
@messense does this seem correct to you? Seems to me like `cargo-zigbuild` might be passing the wrong information down rather than applying a patch here? 🤔

PR #540, opened 2025-08-21
Set PYO3_BUILD_EXTENSION_MODULE env var when building pyo3 extension …

alex:
…modules

See https://github.com/PyO3/pyo3/issues/5202 for context


alex:
Not quite sure what's going on, but I don't think the crossenv failures are related.

alex:
(Confirmed that CI on main is just red ATM :-/)

davidhewitt:
Thanks - I pushed changelog & release commits.

alex:
thank you!

PR #568, opened 2026-02-04
try build emscripten without flags

davidhewitt:
Possible companion to https://github.com/rust-lang/rust/pull/151704


hoodmane:
Okay here's the fix:
https://github.com/rust-lang/rust/pull/152134/
Thanks for catching this @davidhewitt !

hoodmane:
And I confirmed that the nox job successfully builds a while when building against the stage1 compiler with my change. (Didn't run the tests but presumably the wheel actually works too...)

Issue #563, opened 2026-01-29
Options for extracting data files from a `RustExtension`'s build-script `OUT_DIR`


jakelishman:
I have a mixed Python/Rust project using `setuptools-rust` to build a Rust extension module.  This extension module, amongst other things, provides a C API, which I want to expose to external Python extension modules, similarly to how Numpy does (and `rust-numpy` uses) with a function-pointer table stored in a `PyCapsule`.  `setuptools-rust` already solves all my needs for generating the shared-object library extension module, and I can create the `PyCapsule` in PyO3 easily, etc.  However, I want to generate the C header file dynamically during the Rust build (using `cbindgen` as a library), and have the result included in my Python project as a data file, so downstream extension modules can use my Python module as a build dependency.

I think a sensible way I could do this would be if `RustExtension` provides a way for me to specify that a certain list of files or directories should be installed from the `OUT_DIR` of the build script into a specified location within the Python package.  I'm fine if this is limited to a static mapping in `pyproject.toml`.

I'd rather avoid introducing a new build system, or having a bootstrap process before sdist creation; I'm happy with `setuptools-rust`.

I'm happy to produce a prototype PR myself, if this is something that's in scope for the project.

jakelishman:
A complication I hadn't anticipated: a `RustExtension` definition doesn't necessarily correspond one-to-one with a call to `cargo build`.  The most obvious (and problematic) example is macOS universal2 wheels (which I never used, preferring separate Intel and ARM wheels), but the multiple-binary form of `RustExtension.target` has similar behaviour.

This makes it less clear to me how to design a sensible API around those concerns, since I was assuming that the `OUT_DIR` was unique for a single build.

davidhewitt:
MSRV 1.61 is old enough that PyO3 no longer supports it; I think there is machinery for checking the Rust version, we could always emit a friendly error message if the compiler is that old.

PR #572, opened 2026-02-17
Isolate `universal2` artifact-combination logic

jakelishman:
The combination of the multiple build artifacts into one universal2-compatible artifact is now done as part of the general artifact-path handling, which means we no longer have to sort out the configuration environment twice for the same system, and we no longer rely on `cargo` returning information about the built artifacts in the same order between two separate compilations.

Beyond the immediate benefit of isolating the "create binary artifacts" logic into only a single place, this commit is intended to be a refactor along the path towards support for extracting non-binary artifacts that are side-effects of particular `RustExtension`'s build scripts.

---

This supersedes #571, and is a step towards #563 as mentioned (spiritually) in https://github.com/PyO3/setuptools-rust/issues/563#issuecomment-3847256114.

I haven't added any new tests here since it isn't new behaviour, though I couldn't immediately see that the tests in the repo already cover the `universal2` case.  If there's a good way to add them, I can.

I tested it myself locally to verify it produces `universal2` artifacts, and it at least managed to import the resulting object on my ARM Mac (and it's about the right size, and the object functions, etc).

The changes are larger-scale than might be expected because I had to refactor several places of target detection to account for a build process that doesn't necessarily just have a single target triple.

### DIFF
--- a/setuptools_rust/build.py
+++ b/setuptools_rust/build.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import collections
+import enum
 import json
 import os
 import platform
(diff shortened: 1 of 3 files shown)


davidhewitt (approved):
Thanks, these cleanups make sense to me.

The CI has some testing for the universal2 build, looks like that is working ok.

PR #617, opened 2026-09-06
remove `setup.py` from `cffi` example

davidhewitt:
Uses an in-tree build backend for cffi so that we can move fully away from `setup.py` files.


davidhewitt:
Will merge and fix CI in #615

